PR #13, opened 2026-08-08
fix(components): stop file tree from full-rendering and re-rendering all rows [risk:medium]

zxch3n:
**Risk: medium** — file tree interaction (expand/collapse, selection, scroll sync with `ScrollArea`) on every session Files surface and the project file browser. **Confidence: high** on the root cause and the row-count behavior, which are covered by tests each verified to fail without their fix. **Not verified:** the trace-level millisecond claims, which need a production Electron profile on a large repo.

Fixes loro-dev/lody#3339 (the issue lives in the private repo, so no auto-close keyword here).

## Root cause — verified, not assumed

The issue listed a plausible cause; I confirmed the mechanism with a probe test of React's ref attachment order:

```
["layout:null","passive:element"]
```

At the child's `useLayoutEffect` — **the exact pass where TanStack Virtual reads `getScrollElement()`** — the ancestor Radix `ScrollArea` viewport ref is still `null`. React attaches an ancestor host ref only *after* its descendants' layout effects run. So the first virtual range is **always** empty, not just occasionally.

The old code responded to that with:

```ts
if (!shouldVirtualizeRows || virtualItems.length === 0) {
  return rows.map(...) // full list
}
```

which mounted the entire expanded tree at precisely the moment virtualization was needed, then tore it down once a measure succeeded. A viewport that measures 0 kept it there permanently.

## Changes

- **P0 — no full-render fallback.** An unresolved range renders no rows and keeps the total-size spacer, so the scrollport retains the scrollable height a later measure needs. Added `measure()` on viewport attach plus a `ResizeObserver`, mirroring the existing `shared/option-selector.tsx` pattern.
- **P0 — `VirtualFileTreeRow` is `memo`'d.** The virtualizer re-renders the list every scroll frame, which re-ran every mounted row.
- **P0 — icon factories cache by resolved icon name.** A fresh component type per call remounted every icon whenever the parent rebuilt tree data for the same paths, which also defeated the row memo. Keying by icon name bounds the cache by the shipped SVG set rather than by repo file count.
- **P1 — `pruneExpandedFileTreeIds` returns its input Set** when nothing was pruned. File-watcher ticks churn `data` by reference; a fresh equal Set forced an extra render, re-flatten, and virtualizer pass per tick.
- **P1 — one renderer, one threshold.** Removed the tree-wide count (which included collapsed nodes) that swapped Radix `TreeView` in and out; the single gate now counts visible rows.
- **P2** — `renderMode` logs on transitions only, not on every scroll-driven range change. Documented why `w-full` already covers the absolute-row background case (it resolves against the positioned ancestor), so no style change was needed there.

## Verification

Each test was checked against the reintroduced bug rather than trusted for being green:

| Mutation | Result |
|---|---|
| Restore `virtualItems.length === 0` fallback | zero-height test fails — **400 rows mounted instead of 0** |
| Disable `measure()` only | 0 rows render — both halves are load-bearing |
| Un-cache icon factories | 2 identity tests fail |

18/18 pass across the three file-tree suites; `tsc --noEmit` clean.

## Two things to flag

1. `pnpm check:affected` exits 1, but all **26 failures across 8 files reproduce identically on a stashed clean tree** (Convex/platform mock issues in `settings-data-cache`, `useOrganization`, `workspace-target-router`, and others). Pre-existing and unrelated — none touch these files.
2. `pnpm format` reformatted 18 unrelated CLI/Electron files that were already unformatted at the merge-base; I reverted those to keep the diff surgical.

## Story fidelity

`FileTreeList.stories.tsx` claimed to mirror `FileTreeView`'s renderer while composing `TreeView` directly — this change made that false. It now renders the real `FileTreeProviderView` per the Storybook-fidelity invariant, and adds a 600-row story for the windowed path.

## Manual check still wanted

Electron session Files panel on a large repo: expand/collapse, selection, and fast scroll.

### DIFF
--- a/packages/components/src/components/icons/file-icons/index.tsx
+++ b/packages/components/src/components/icons/file-icons/index.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useMemo, type ComponentType } from 'react';
 import {
   compoundExtensionMap,
   extensionMap,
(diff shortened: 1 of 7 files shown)


zxch3n:
## Follow-up: restored the truncated-name tooltip (543462e)

Routing small trees through `VirtualFileTree` silently changed row label behavior — I missed this in the original push and want it on the record rather than buried in a diff.

The shared `TreeView` row this renderer replaced showed the **bare file name** in a styled Radix tooltip, and **only when the label was actually clipped** (`scrollWidth > clientWidth`). The virtual row used `title={item.id}`, which meant:

- browser-native yellow box instead of the app's tooltip surface
- the **full path** instead of the name
- fired even for names that fit

Restored via `VirtualFileTreeRowLabel`, plus the `TooltipProvider` ancestor Radix requires (`TreeView` supplied its own; these rows had none).

**One deliberate difference from `TreeView`:** truncation is measured on **hover**, not in a per-row mount effect. Reading `scrollWidth` from an effect is fine in a static list, but in a virtual list it forces a layout for every row scrolled into view — which would give back part of the win this PR exists for. Hover is also the only moment the answer is needed.

Both new tests were verified to fail against the `title` version:

| Mutation | Result |
|---|---|
| Revert label to `title={item.id}` | `does not put a native title tooltip on row labels` fails; `shows a name-only tooltip…` fails |

20/20 now pass across the three file-tree suites; `tsc --noEmit` clean.

### Also verified since opening this PR

- Private `apps/web` / `apps/mobile` do not reference the removed `shouldVirtualizeFileTreeData` / `countTreeDataItems` exports — no build break outside this repo.
- `packages/code-review-helper`'s standalone icon copy has no factory functions, so nothing to sync there.
- **Threshold crossing** (the case I had not tested before opening this PR): a lazy directory that is 1 row collapsed and 301 expanded. `enabled: false` clears the virtualizer's measurement cache, so this was the plausible break. After expanding: 22 rows mounted, spacer 6622px (= 301 × 22) — virtualization engages correctly.

Still unverified, as stated in the description: the issue's millisecond-level trace numbers, which need a production Electron profile on a large repo.

PR #41, opened 2026-08-26
fix: resolve TODO comments in json-lines-parser and root route

clanzhang:
## Summary

Resolve 2 TODO comments found in the codebase.

### Changes

**`apps/cli/src/utils/json-lines-parser.ts`**
- Replace silent `catch` block with `this.emit('warning', ...)` so malformed JSON lines are observable by consumers instead of being swallowed

**`packages/components/src/routes/__root.tsx`**
- Add `title`, `description`, `theme-color`, and Open Graph meta tags to the root route `head()` — previously only `charset` and `viewport` were set

### DIFF
--- a/apps/cli/src/utils/json-lines-parser.ts
+++ b/apps/cli/src/utils/json-lines-parser.ts
@@ -29,8 +29,10 @@ export class JsonLinesParser extends Transform {
           const parsed = JSON.parse(line);
           this.push(parsed);
         } catch {
-          // TODO: better parse error
-          // console.log("#### error", line);
+          this.emit(
+            'warning',
+            new Error(`Skipping malformed JSON line: ${line.slice(0, 200)}`)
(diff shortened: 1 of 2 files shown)


apps/cli/src/utils/json-lines-parser.ts
```diff
@@ -29,8 +29,10 @@ export class JsonLinesParser extends Transform {
           const parsed = JSON.parse(line);
           this.push(parsed);
         } catch {
-          // TODO: better parse error
-          // console.log("#### error", line);
+          this.emit(
```
Leeeon233, line 32: I can't remember if there might be a problem here. I'm inclined not to modify this for now.

clanzhang: Thanks for the review! I'll close this PR.

PR #49, opened 2026-08-27
fix: initialize loro metadata monitoring

hyoban:
## Summary

- patch `loro-repo@0.20.0` so its metadata live monitor starts during repository initialization
- keep transport-free local meta imports hydrated in the repo metadata cache and visible to `repo.watch({ kinds: ['doc-metadata'] })`
- remove the regression tests' manual activation of the private sync runner and document the local dispatch invariant

## User impact

In the OSS Electron app, a locally sent message could appear in the conversation and remain durably persisted while the agent never started and no assistant reply arrived. Restarting the desktop app could recover the pending turn through the startup session scan, but that was only a temporary recovery: a later live message could become stuck again.

## Root cause

The OSS Electron composition starts the CLI repo without a registered transport and syncs renderer metadata through the local Loro data plane. The renderer's `latestUserMsgId` update reached the CLI's raw meta Flock, but `loro-repo` only started `ensureMetaLiveMonitor()` during metadata sync or an actual meta transport attachment. A pending transport-free meta room therefore never hydrated the derived metadata cache or emitted the event consumed by `SessionDispatchWatcher`.

Starting the existing idempotent monitor from `LoroRepo.ready()` closes that initialization gap without registering a transport or making a cloud request. Later cloud sync and transport attachment calls remain safe no-ops for monitor setup.

## Why a dependency patch

The published package metadata points to `loro-dev/loro-repo`, but that repository is currently unavailable to this contributor, and the public `loro-dev` repositories do not contain this package's source. There is therefore no accessible upstream target for a source PR today.

This change carries the smallest reproducible fix through pnpm's version-pinned `patchedDependencies`. When the upstream source becomes available and publishes the initialization fix, Lody should upgrade `loro-repo` and remove this patch.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm exec vitest run tests/doc-meta-subscription.test.ts` (7 passed)
- `pnpm exec vitest run tests/session-dispatch-watcher.test.ts` (36 passed)
- `pnpm exec oxlint --type-aware --quiet packages/components/tests/doc-meta-subscription.test.ts`
- `pnpm check:platform-boundaries`
- `pnpm check:public-boundary`
- `pnpm start:local` (desktop rebuilt and launched; a persisted pending turn was recovered and completed)

`pnpm check` completed the workspace typecheck, then stopped in the repository's existing lint baseline: 68 errors, primarily missing type definitions in the isolated `packages/acp-extension-kimi` submodule, plus an unrelated unused parameter in `packages/components/tests/session-file-content-view.test.tsx`.

This PR is intended as a minimal downstream fix while the package source is unavailable. If the maintainers prefer to address the issue through an upstream release or a different implementation, please feel free to close or supersede this PR.

### DIFF
--- a/apps/cli/src/lib/loro/AGENTS.md
+++ b/apps/cli/src/lib/loro/AGENTS.md
@@ -12,6 +12,11 @@ down. Cost per call = one Streams subscription plus the doc's full initial sync.
 
 Rules:
 
+- Renderer metadata reaches the CLI by direct import into the repo's internal
+  meta Flock even when local mode has no registered transport. Keep the
+  `loro-repo` metadata live monitor enabled from repo initialization; deferring
+  it until transport join leaves `getDocMeta` stale and prevents the session
+  dispatch watcher from seeing `latestUserMsgId`.
 - Never open docs in a loop over `listAliveRoomIds` or any other workspace-wide
   enumeration. A long-lived workspace holds thousands of historical session
   rooms; opening them all stalls startup and floods the Streams backend.
--- a/packages/components/tests/doc-meta-subscription.test.ts
+++ b/packages/components/tests/doc-meta-subscription.test.ts
@@ -235,7 +235,9 @@ describe('docMetaSubscriptionAtom', () => {
     const machineId = 'machine-1' as MachineId;
     const docId = getMachineRoomId(machineId);
 
-    await repo.syncRunner.ensureMetaLiveMonitor();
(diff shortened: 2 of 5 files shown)


Leeeon233:
@zxch3n take a look please

zxch3n:
LGTM. I'll merge this first then fix the upstream issue

PR #52, opened 2026-08-27
fix(components): constrain message editor in narrow columns

hyoban:
## Summary

- use the available message-column width for both wrappers while editing
- preserve the existing fit-content message sizing outside edit mode
- keep the editor border and actions visible when a session side panel narrows the conversation

## Before / after

| Before | After |
| --- | --- |
| Message editor overflowing a narrow conversation column before the fix | Message editor contained within a narrower conversation column after the fix |

## Validation

- `corepack pnpm --filter @lody/components typecheck`
- `corepack pnpm --filter @lody/components test` (380 files, 2727 tests)
- `git diff --check`
- manually verified in the local Electron app with HMR and the session side panel open

## Full check note

`corepack pnpm check` completed workspace typechecking, then stopped in lint on unrelated existing errors, including `packages/components/tests/session-file-content-view.test.tsx` and missing type definitions in the isolated Kimi workspace.


Leeeon233:
Thanks

PR #62, opened 2026-08-27
fix(components): show latest usage weeks on compact calendars

hcz-czh:
## Summary

- initialize the usage heatmap at its right edge when the calendar overflows
- keep the most recent activity visible by default in the 30-day and all-time ranges
- apply the initial position only on mount so subsequent user scrolling is preserved

## Verification

- `pnpm --filter @lody/components typecheck`
- `pnpm --filter @lody/components exec vitest run tests/usage-calendar-model.test.ts` (9 tests)
- targeted Prettier and Oxlint checks on the changed component
- `git diff --check`
- verified the 30-day and all-time Storybook stories initialize at the right edge at 640px
- verified the calendar has no extra horizontal scrolling at 1280px


wibus-wee (approved):

PR #71, opened 2026-08-27
fix(electron): block untrusted main-window navigation

Pleasurecruise:
## Author type

- [ ] I am a human

## Problem / pressure

The privileged Electron main window could navigate to a remote document while retaining its preload bridge, allowing remote content to reach high-privilege IPC and ultimately execute local commands.

## Summary

Restrict top-level main-window navigations and redirects to the existing application and recovery renderer targets. Blocked HTTP(S) destinations are opened in the system browser.

## Before / after

| Before | After |
| ------ | ----- |
| Remote content could replace the main renderer and retain privileged preload access. | The main window prevents untrusted navigation before commit and keeps only application-owned targets. |

## Test plan

- Ran Electron main/preload TypeScript checking successfully.
- Ran the existing Electron test suite: 43 tests passed.
- Ran `git diff --check` successfully.

- **Review focus:** Inspect `apps/electron/src/main/window.ts`, especially URL normalization and both navigation event handlers.
- **Decisions to challenge:** Confirm same-origin development navigation and exact packaged-file matching are the intended trust boundaries.
- **Plausible failures / evidence gaps:** No dedicated navigation test or manual packaged Electron smoke test was added; verification used type checking and the existing suite.

### Authoring context

- **User goal / directives:** Fix remote navigation retaining privileged Electron IPC, use an isolated worktree, keep names concise, and create a PR.
- **Constraints / non-goals:** Keep the change narrow; do not add test files, documentation changes, preload changes, or broad IPC refactors.
- **Risk-bearing decisions:** Reuse the resolved application and recovery targets as the allowlist, compare development targets by origin, and open blocked HTTP(S) URLs externally.
- **Destructive or irreversible behavior:** None; the change only prevents unsafe navigation and redirects HTTP(S) destinations to the system browser.
- **Deliberately not done or tested:** No new `.mjs` test and no manual packaged-app smoke test, per the requested narrow scope.
- **Unknowns / confidence:** High confidence in Electron 39 event typing and existing test coverage; packaged navigation behavior was not exercised manually.

### DIFF
--- a/apps/electron/src/main/window.ts
+++ b/apps/electron/src/main/window.ts
@@ -1,6 +1,7 @@
 import { app, BrowserWindow, dialog, nativeTheme, shell } from 'electron'
 import { is } from '@electron-toolkit/utils'
 import { join } from 'node:path'
+import { pathToFileURL } from 'node:url'
 import {
   consumePendingDeepLink,
   getMainWindow,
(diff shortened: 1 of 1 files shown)


wibus-wee (commented):

PR #86, opened 2026-08-27
feat(components): show activity and unread results on collapsed projects

### LINKED ISSUE #137: [Feature Request] Show activity and unread results on collapsed projects
### Affected area

Desktop app

This feature request comes from a small need I encountered while using Lody in my own day-to-day workflow.

I often keep projects collapsed to reduce sidebar noise and make it easier to focus on the project I am currently working in. However, once a project was collapsed, its Session rows and their status indicators were hidden as well. I could no longer tell at a glance whether the project still had a running task, was waiting for permission, or contained a completed result I had not read.

This makes it difficult to monitor several projects while keeping the sidebar compact.

### Desired outcome

When a local project or GitHub repository group is collapsed, I would like to see a compact summary of relevant Session activity on the project row.

When the group is expanded, the existing individual Session indicators should remain the source of detail and the project-level summary should disappear.

The goal is to preserve visibility of important activity without requiring projects to remain expanded.

### Proposed solution

A possible approach is to reuse the existing Session row indicator and aggregate the states of Sessions belonging to the collapsed group.

The aggregate could cover:

- Local projects, including projects running on remote machines.
- GitHub repository groups.
- Pinned Sessions and hidden child Tabs.
- Running, initializing, permission-request, and unread completed-result states.

Permission requests should have the highest display priority. Running and initializing can reuse the existing spinner with different accessible descriptions. If active work and a separate unread result coexist, the spinner could include a centered solid dot to communicate both states.

The aggregate indicator should only be shown while the project or repository group is collapsed.

The current workaround is to keep every project expanded or open each project periodically to inspect its Sessions. Keeping projects expanded increases sidebar noise, while checking each project manually makes it easy to miss running work or unread results.

### Trade-offs or non-goals

This should remain a focused sidebar usability improvement.

It should not redesign the existing Session status model, change Session persistence, add a new notification system, or alter the behavior of individual Session rows.

The exact visual treatment and state priority can be adjusted based on maintainer feedback.

### Additional context

This request is based on a real usage need: I keep projects collapsed to reduce sidebar noise, but still need to know whether hidden work is running or whether a completed result requires attention.

A draft implementation exists in [PR #86](https://github.com/LodyAI/Lody/pull/86) to make the proposed behavior concrete. The implementation should not be considered maintainer-approved; this Issue is intended to request agreement on the scope and approach.

### Before submitting

- [x] I will wait for a Lody maintainer to explicitly agree on the scope and approach before implementation or a pull request.

hcz-czh:
## Related issue

Closes #137

## Problem / pressure

When projects are collapsed, their Session rows are hidden, so users cannot tell whether the project contains a permission request, active work, or unread results.

This feature came from a small need encountered during my own use of Lody. I often keep projects collapsed to reduce sidebar noise, but then could not tell whether a project still required attention. The project row should provide a compact navigation signal without duplicating the detailed status indicators shown when the project is expanded.

## Summary

Add a two-level activity summary to collapsed local-project and GitHub repository rows.

The summary:

- Aggregates fresh activity from Sessions and child Tabs.
- Includes pinned Sessions in local-project summaries.
- Supports local projects, remote-machine projects, and GitHub repository groups.
- Displays up to two status levels in the trailing area of collapsed project rows.
- Uses project-level priority `permission > unread > active`.
- Treats `active` as higher priority than `unread` when the same Session has both states and all unread Sessions are also active.
- Counts unique Sessions rather than status occurrences.
- Excludes Sessions already represented by the primary item from the secondary item.
- Shows a remaining state directly when only one state type remains.
- Combines multiple remaining state types into a deduplicated `+N` count.
- Shows count `1` when needed to clarify a mixed `+N` aggregate.
- Omits ordinary counts when the count is one.
- Uses the neutral `Active` label for running and initializing Sessions.
- Places a single summary item in the right slot and aligns two summary items across fixed slots.
- Hides the project-level summary when the project or repository is expanded.
- Preserves existing Session indicators, opened-by tree behavior, disclosure controls, and row actions.

## Visual explanation

```text
Collapsed project/repository
  Sessions + child Tabs
        |
        v
  unique Session activity
        |
        +-- primary: highest-priority project state
        |     permission > unread > active
        |
        +-- secondary:
              no remaining Session -> hidden
              one remaining type   -> its icon and count
              multiple types       -> deduplicated +N
```

Examples:

```text
A: permission
B: running + unread
→ ✋ 1   +1
```

```text
A: unread
B: active
→ ●      ↻
```

```text
A: running + unread
→ ↻
```

When the project is expanded, the aggregate summary is hidden and the individual Session rows remain visible.

## Before / after

| Before | After |
| ------ | ----- |
| Collapsing a project hid its Session activity. | A collapsed project shows a compact activity summary. |
| Multiple permission requests were represented by one icon. | Permission requests display a count when greater than one. |
| A Session with multiple states could inflate the secondary count. | Secondary counts use unique remaining Session IDs. |
| GitHub repository rows did not summarize hidden Session activity. | Collapsed GitHub repository groups use the same summary model. |
| Expanded projects could show redundant project-level information. | Expanded projects hide the aggregate while Session rows retain their indicators. |

## Test plan

- Ran the project activity test suite: 60 tests passed.
- Ran the components TypeScript typecheck successfully.
- Ran formatting and `git diff --check` successfully.
- Added coverage for a Session that is both running and unread.
- Verified deduplicated secondary counts and mixed `+N` rendering.
- Verified single-item right alignment and fixed two-slot layout.
- Updated the `Collapsed Project Activity` Storybook to use the real sidebar layout.
- Storybook covers representative local-project, remote-project, and GitHub Worktree states, including permission, unread, active, initializing, overlapping, and mixed activity.
- Verified that existing row actions and disclosure controls remain available.

- **Review focus:** Check `project-activity.ts`, `project-activity-indicator.tsx`, `session-list-rows.ts`, `loro-app-sidebar.tsx`, and `session-list.tsx`.
- **Decisions to challenge:** Verify the unique-Session meaning of `+N`, same-Session active and unread handling, project-level priority, neutral `Active` labeling, and fixed-slot placement.
- **Plausible failures / evidence gaps:** Check whether overlapping Sessions, child Tabs, pinned Sessions, local projects, remote projects, and GitHub repository groups all produce consistent collapsed-row summaries.

### Authoring context

- **User goal / directives:** Keep meaningful project activity visible while local and GitHub projects remain collapsed.
- **Constraints / non-goals:** Preserve existing Session status semantics, opened-by tree behavior, project actions, and expanded-row behavior. Do not redesign the Session state machine.
- **Risk-bearing decisions:** Permission requests remain the highest-priority project state. A Session with both active and unread state is counted once. The secondary `+N` value represents unique remaining Sessions.
- **Destructive or irreversible behavior:** None. This change only affects presentation and deterministic aggregation.
- **Deliberately not done or tested:** No broader redesign of Session status semantics or persistence behavior was attempted.
- **Unknowns / confidence:** Components tests, typecheck, formatting checks, and representative Storybook scenarios passed. Maintainers may still prefer a different visual treatment for the two summary slots.

### Original user prompt

<details>

````text
第一轮：查看当前Lody是否存在将项目折叠起来之后，若项目下仍然有在执行的任务时，项目条目上也有像session执行那样类似的提示
第二轮：那我就要做一下这个功能然后给它提PR了，但是运行状态的显示也有是需要考虑的，是直接参考session的方式，覆盖文件夹图标来显示运行状态还是有其它更好的选择呢？
````

</details>

### CODE REVIEW

**Comment by hcz-czh** [comment]
@wibus-wee Could you review this PR when you have a chance? It adds activity and unread-result indicators to collapsed projects and repository groups, based on a small need I encountered while using Lody.

I'd appreciate your thoughts on whether this approach fits the sidebar. If the direction looks reasonable but the visual treatment or status priority needs adjustment, please leave a comment. Thank you!

**Comment by wibus-wee** [comment]
@hcz-czh First of all, thanks for the contribution!

I've been thinking about this, and I agree that this is a UX problem. But I think the hardest part is that once we start aggregating multiple sessions, we can no longer treat this as simply picking one `SessionStatus` to display.

The current approach in the PR is basically:

`permission > running > initializing`

with `unread` layered on top as a separate boolean.

This is straightforward to implement with our current code, but from a UX perspective, it makes an implicit assumption: **the user only needs to know about the single most important state.** That assumption starts to break down once multiple sessions are involved.

These states don't belong to the same state machine, so I don't think we should force them to override one another.

For example, if the collapsed state only shows:

```
Lody                          ✋
```

that's technically correct, but it loses quite a bit of information. The user knows that something needs permission, but has no idea that other sessions are still running or that there are two unread results waiting for them.

So I think the most reasonable model is a **two-level aggregate at most**. For example:

```
Lody                      ✋ 1   +3
```

The first indicator always communicates the highest-value piece of information, while the second simply tells the user that there's more going on.

In other words, I think we should give the sidebar an explicit **information budget** rather than trying to compress every session into a single status.

**Comment by hcz-czh** [comment]
@wibus-wee 

Thank you for your reply!

I understand the motivation behind the proposed two-level information budget. When a project contains multiple Sessions with different states, simply selecting one highest-priority state can hide other relevant information.

Based on my own usage, the main purpose of the project row is not to provide a complete summary of every Session. It is to offer a low-cost overview while the project is collapsed. When a project contains a Session waiting for permission, users usually have a strong reason to open it and handle the request immediately. For that reason, I think permission requests should remain visually prominent.

For unread results and other activity, I think the usefulness of an aggregate indicator may depend on whether the project is expanded:

- When the project is collapsed, a compact aggregate indicator helps users notice that something requires attention without opening the project.
- When the project is expanded and contains only a few Sessions, users can already see each Session’s state directly, so an additional project-level summary may not be necessary.
- When the project is expanded but contains many Sessions, a count on the project row could still be useful because users may need to scroll through the list. At the same time, that count may be easy to miss once the user is focused on the Session rows, so the practical benefit may be limited.

My current design focuses on providing an overall signal without adding multiple status icons. Once users see that a project contains activity or unread results, they can expand it and inspect the individual Sessions. Because of this, I am not sure that showing counts for every state on the project row is necessary, especially if doing so increases the user’s visual and cognitive load.

I also have a slightly different perspective on the idea that aggregating multiple Sessions means we can no longer select the most important `SessionStatus` to display. From my perspective, the collapsed project indicator is primarily a navigation signal rather than a complete status report. It only needs to identify the most important condition and indicate that there may be additional items to review. The complete information remains available after expanding the project.

That said, this view is based mainly on my own experience using Lody. Perhaps a broader round of user research would help clarify which approach is most useful. If we can agree on a clear interaction model, I would be happy to adjust the implementation accordingly.

**Comment by wibus-wee** [comment]
@hcz-czh IMO, Consider a project with:

* Session A: `permission`
* Session B: `running`
* Session C: `unread`

If the collapsed state only shows:

```
Lody    ✋
```

what the UI is actually communicating is “This project has a permission request.”
rather than “This project needs your attention, and there is other activity happening as well.”

Those are meaningfully different pieces of information.

With the current interaction model, the meaning of the indicator also becomes inconsistent. Sometimes the center dot represents an unread result, while in other cases `permission` takes precedence and masks every other state.

From an interaction-design perspective, I also don't think `loading + center dot` forms an intuitive or easily understandable combination. There isn't an obvious visual grammar that tells the user what those two indicators mean together.

**Comment by hcz-czh** [comment]
@wibus-wee 

Now I understand that a two-level indicator can distinguish between “there is a permission request” and “there is a permission request alongside other activity.” That does preserve more information than the current approach.

I'm still unsure how much the additional count would help users decide what to do next, though, since it doesn't indicate what those other activities are. I don't have a more satisfying alternative at the moment either.

I also agree that the spinner with a center dot may need some explanation rather than being immediately understandable, so that part is worth reconsidering.

**Comment by wibus-wee** [comment]
@hcz-czh Thanks for the thoughtful reply. I think we're converging on the model, so let me sharpen the remaining argument and then talk about presentation.

On "the count doesn't say what those other activities are": you're right that a bare count is weak information. But the count's job isn't to describe the activities — it's to calibrate the user's decision. ✋ 1 +3 and ✋ 1 lead to different actions: with the former, the user expands the project knowing there's a backlog; with the latter, they handle the permission request and move on. A navigation signal doesn't need to say what the rest is; it needs to say how much attention is waiting.

The stronger reason I keep coming back to two levels, though, is urgency masking. permission requires an immediate response; unread can wait. When a single highest-priority state completely hides the rest, a second
permission request — equally urgent — becomes invisible until the user happens to expand the project. The failure isn't that we picked one status over another; it's that states with different urgency silently delay
each other. Aggregating low-urgency noise into a count is fine — that half of your "navigation signal" position I fully agree with. Hiding a second urgent state is not.

And on consistency: with the current approach, the same visual element (the center dot) means "unread" in some contexts and is masked by permission in others. The user can't build a stable mental model of the
indicator. With two levels, the reading rule is constant: "the highest-value event, plus more." That stability is worth more than the information in the numbers themselves.

For the presentation, I agree the spinner + center dot combination isn't self-explanatory. Rather than debate it abstractly, let me propose a concrete direction for the collapsed row:

```
  Lody ✋ 1   +3
```

- first slot: always the single highest-urgency state, with its count if > 1 (✋ 2 for two permission requests)
- second slot: a quiet +N aggregating everything else (running, unread, initializing)

If the visuals still feel unclear, I'm open to alternatives — e.g. replacing +N with a small dot cluster — as long as the two-level rule stays. :)

**Comment by hcz-czh** [comment]
@wibus-wee 

Thanks! 

I agree that the count is important because it can affect whether the user decides to open the project. Knowing that there are three items waiting for attention is different from knowing that there are ten.

I think we could define the two levels dynamically:

- The first level shows the highest-priority state and its count, omitting the count when it is one.
- If only one other state remains, the second level shows that state and its count.
- If multiple other states remain, the second level combines them into a single total count.
- If only one state exists, only the first level is shown.

For example:

```text
Lody    ✋
```

One permission request.

```text
Lody    ✋ 2   ● 3
```

Two permission requests and three unread results.

```text
Lody    ✋ 1   +5
```

One permission request and five other Sessions with multiple different states.

```text
Lody    ● 3   ↻ 2
```

Three unread results and two active Sessions.

```text
Lody    ↻
```

One active Session.

This keeps the first level focused on the most urgent state. It also preserves the type of the remaining state when there is only one, while using a compact total when several lower-priority states coexist. A tooltip could provide the detailed breakdown if needed.

wibus-wee:
Hi, sorry for the long wait :)

I will review your PR again.

hcz-czh:
@wibus-wee 

No worries—good things take time :) I just thought of one detail regarding the mixed aggregate case.

If we render `✋ +5`, the ownership of the numbers may be ambiguous: it could mean one permission request plus five items in other states, but it might also be read as five additional permission requests.

To make this clearer, perhaps we could omit a count of `1` in the usual cases, but show it explicitly when the second level is a mixed `+N` aggregate:

```text
✋            one permission request
✋     ● 3    one permission request and three unread results
✋ 1   +5     one permission request and five items across the remaining states
```

This would keep the common cases lightweight while making the number ownership explicit in the otherwise ambiguous mixed-state case. What do you think?

wibus-wee:
it's a good idea! @hcz-czh 

But i think the current `+N` is not actually counting “other Sessions”. It is counting remaining status occurrences.

For example, consider:

* Session A: permission
* Session B: running + unread

There are only two Sessions, and only one other Session besides the permission one. But because Session B contributes to both `unread` and `active`, the collapsed row becomes:

```text
✋ 1   +2
```

I think most users will naturally read `+2` as “two other things/Sessions need attention”, not “one Session currently contributes two status dimensions”.

This matters because the main UX justification for the second-level count was that it helps the user calibrate whether expanding the project is worthwhile. If one Session can increase that number twice, the number stops being a reliable navigation signal.

The spec currently explicitly says the count represents status occurrences rather than deduplicated Sessions, but I think that semantic is too implementation-shaped for the UI.

My preference would be:

```text
primary = highest-priority state + number of Sessions in that state
secondary = number of remaining unique Sessions not represented by the primary slot
```

So the example above would be:

```text
✋ 1   +1
```

There is also a smaller accessibility/labeling issue: `running` and `initializing` are merged into `active`, but the tooltip / aria label currently calls all of them “Running”. An initializing-only project therefore gets announced as “Running”. If those states intentionally share one aggregate bucket, I think the aggregate label should be something neutral such as “Active”.

The existing tests and Storybook cases don't catch the first issue because their fixtures create separate Sessions for `permission`, `unread`, and `active`; they don't exercise a real Session that is both active and unread.

Before merging, I'd like us to explicitly decide what unit `+N` represents and add at least one overlapping-state fixture (`running + unread` on the same Session). Once that semantic is stable, the rest of the presentation looks much easier to evaluate.

hcz-czh:
@wibus-wee 

Thanks for pointing this out. I agree that `+N` should represent unique remaining Sessions rather than status occurrences.

I updated the collapsed project aggregation:

* Each Session contributes at most once to the displayed project-level summary.
* When the same Session is both running and unread, `active` takes precedence when it is the only project-level state.
* Across different Sessions, the project row prioritizes `permission > unread > active`.
* Sessions represented by the primary item are excluded from the secondary item.
* The secondary `+N` count is based on unique remaining Session IDs, so one Session cannot inflate the count by contributing multiple status dimensions.
* `running` and `initializing` use the neutral `Active` label in aggregate tooltips and accessibility text.
* Single-item summaries are placed in the right slot, while two-item summaries use both fixed slots.

Examples:

```text
A: permission
B: running + unread
→ ✋ 1   +1
```

```text
A: unread
B: active
→ ●      ↻
```

```text
A: running + unread
→ ↻
```

I also added an overlapping `running + unread` fixture and updated the Storybook coverage to verify the deduplicated count and collapsed-row presentation.

The PR description has also been updated.

.oxlintrc.json
wibus-wee, line 1: something unrelated in your diff

packages/components/tests/project-activity.test.tsx
wibus-wee, line 1: Remove UI unit test

packages/components/tests/session-list-opened-by.test.ts
```diff
@@ -185,6 +189,17 @@ describe('SessionList opened-by rendering', () => {
     return { onSelectSession, store };
   }
 
+  it('hides repository aggregation while the individual Session rows are expanded', () => {
```
wibus-wee: Unnecessary test.

wibus-wee (changes requested):
I think the implementation is in a much better place now, but the PR has accumulated quite a bit of validation/documentation scaffolding along the way.

Could we trim some of it before merging? In particular, some behavior is now covered in multiple places, and a few tests assert DOM/layout implementation details rather than user-visible behavior.

I'd prefer keeping one source of truth for the aggregation rules, focused tests for the important edge cases, and a small representative Storybook set. The feature itself is relatively small, so I don't think we need to permanently carry every intermediate review case in the repo.

.oxlintrc.json
hcz-czh: Removed the unrelated `.oxlintrc.json` change. The PR no longer modifies this file. Addressed in `d717343d`.

packages/components/tests/project-activity.test.tsx
hcz-czh: Replaced the UI/DOM-focused tests with a small set of pure aggregation tests. The remaining cases cover status priority, overlapping `running + unread` Sessions, unique-Session counting for `+N`, primary-slot exclusion, and duplicate/nested input deduplication. Addressed in `d717343d`.

packages/components/tests/session-list-opened-by.test.ts
hcz-czh: Removed this test and reverted the related helper changes. This file is now unchanged from the upstream base. Addressed in `d717343d`.

hcz-czh:
@wibus-wee 

Thanks for the review. I trimmed the validation and documentation scaffolding in `d717343d`:

* removed the unrelated `.oxlintrc.json` change;
* replaced the UI/DOM implementation-detail tests with focused pure aggregation tests;
* removed the unrelated `session-list-opened-by` test;
* reduced Storybook to one representative story;
* condensed the contributor guidance while retaining the aggregation contract.

The aggregation behavior itself is unchanged, and the focused tests, type checking, `pnpm check:quick`, formatting checks, and Storybook build all pass.

wibus-wee:
LGTM, could u take a screenshot for me?   @hcz-czh

hcz-czh:
@wibus-wee 

Sharing the screenshots for the two approaches here 🙂

**Image 1 — current implementation**

A

Each collapsed project row keeps space on the right for two hover-only actions: the “More actions” menu (`MoreHorizontal`, ⋯) and the “New chat in this project” button (`SquarePen`, ✎).

The two activity slots sit just to the left of that action area, so they stay aligned and don’t move when the icons appear. The trade-off is that, when the row isn’t hovered, the reserved action space leaves a little extra blank space on the right.

**Image 2 — alternative treatment**

B

In this version, the activity slots sit flush against the right edge at rest, using the space the actions would otherwise reserve.

On hover, the action area expands and shifts the activity slots to the left, revealing the ⋯ and new-chat buttons without overlapping the status.

So in short:

* at rest — activity slots sit flush right, with no extra reserved space;
* on hover — the actions expand in and the slots shift left.

I also included the row tooltip and expanded-group states in the screenshots.

Both approaches keep the two activity slots aligned across project and repository rows. I’m happy with either direction — the current version is a bit more stable, while the second is a bit tighter at rest. Curious which one feels better to you 🙂

wibus-wee:
Thanks for putting both approaches together. Seeing them side by side actually makes me wonder if we should avoid adding aggregated activity to project rows altogether.

The right side of the row is already carrying quite a lot. Once we add aggregated status, counts, hover actions, and potentially more states later, the row starts to feel more like a monitoring surface than a navigation item. Both approaches handle the layout reasonably well, but I’m not sure the underlying complexity should live in the row in the first place.

If we eventually introduce filtering / sorting by status — something similar to the status-based views in Codex — that might be a cleaner way to surface this information. When I want to see what needs permission, has unread activity, or is currently running, the filtered list itself becomes the answer instead of every project row having to encode all of those states at once.

In that direction, I’d prefer keeping project/repo rows visually simple, and letting status primarily drive filtering / ordering. We could still keep a very lightweight indicator for something truly actionable, like a permission request or something else, but I’m not convinced we need the full permission + unread + active aggregation on every row.

So I’m wondering if the better solution is not really A vs. B, but avoiding this pattern entirely once we have a stronger status-based filtering model.

hcz-czh:
@wibus-wee 

I share that concern — seeing both layouts side by side raised the same doubt for me. For what it's worth, here's how the two mechanisms compare from my side:

A row aggregate and a status filter answer different questions. The filter answers "what needs me": it aggregates across every project in one place, costs the row nothing, and scales by adding one more filter option instead of renegotiating a fixed slot budget. On those terms the row clearly loses, and the navigation-vs-monitoring framing is right.

The one structural difference is that a filtered view is pull-based — it produces a signal only when the user opens it. The row aggregate is push-based — it works while the user isn't looking. Today that second case has no other in-app surface: the permission card renders only inside the open session, and the OS notification fires only while the app is backgrounded (and only at the moment of the transition — a request that arrives while the app is focused never triggers it). So with rows carrying nothing, a permission request inside a collapsed project stays invisible until the user already suspects it and goes looking. That's the residual case the filtering model would leave open — it may well be acceptable, or better served by extending notifications to the foreground than by anything on the row.

By the way, some history: this PR actually started out as roughly that lightweight indicator — the first version was just a single status mark on collapsed rows, and the count/+N aggregation only grew later as it tried to cover more states. The PR was always meant as a starting point to open up this discussion rather than a settled design.

Either way, I'm not attached to A or B — they were just two treatments of the layout, and I agree the underlying pattern is the bigger question. If the stronger filtering model feels like the better direction, with the existing notification path covering the push case — perhaps extended to foreground transitions if needed — or with the lightweight actionable indicator you mentioned, I'm happy to treat this PR as exploration rather than something we need to land. We can close it or park it for now and revisit once that direction is clearer, whichever makes the most sense.

wibus-wee:
@hcz-czh Thank you for your understanding. I believe we need to discuss this internally. We will be rolling out a series of UI/UX updates soon. If any changes relate to this PR, I will list you as a co-author. I appreciate the thorough discussion we’ve had. Thank you.

hcz-czh:
@wibus-wee 

Sounds good. I’ve learned a lot through these discussions as well, and I really appreciate the time you’ve taken to talk through the design with me. Lody is a great product — I genuinely love using it — so I’m glad I could contribute to the discussion.

I mostly saw this PR as a way to explore the problem and get the conversation started, so I’m not particularly attached to the PR itself. I’ll move it back to draft for now and leave the broader direction to your internal UI/UX discussion.

Looking forward to seeing where Lody goes next, and thanks again for the thoughtful discussion.

PR #87, opened 2026-08-27
fix(components): reapply project selection from sidebar

sheepbox8646:
## Related issue

https://github.com/LodyAI/Lody/issues/153

The Issue was created after this PR to correct an authoring-process oversight and is awaiting explicit maintainer agreement.

## Problem / pressure

Clicking a local-project row is the primary way to select that project for the Composer. `ChatLanding` applies a URL pre-selection once per intent and then lets the user steer freely, while clearing the project in the Composer leaves the URL unchanged. If the user then clicks the same project row again, byte-identical search state previously made the click a no-op instead of re-applying the visible project.

## Summary

- Keep the project row as the single primary selection interaction; remove the overlapping hover-only new-Session control.
- Add a fresh `projectSelection` intent to every project-row navigation and include it in `buildChatLandingPreSelectionKey`.
- Generate intent keys from the timestamp plus a monotonic sequence so repeated clicks remain distinct even within the same millisecond.
- Preserve the typed draft instead of reusing `resetDraftKey`, and remove the obsolete prop, story, icon, and translation wiring.

## Before / after

| Before | After |
| ------ | ----- |
| Clicking the project already named by the URL could not re-select it after the Composer selection was cleared. | Every project-row click re-applies that project as a fresh Composer intent. |
| The proposed hover action duplicated the row's selection purpose. | The row remains the sole selection interaction; remove stays the only secondary row action. |

## Test plan

- `mise exec node@22 -- pnpm --filter @lody/components typecheck` — passed.
- `mise exec node@22 -- pnpm --filter @lody/components test` — 392 files / 2798 tests passed, including distinct same-millisecond project-selection intents.
- Prettier on all changed files, `git diff --check`, `mise exec node@22 -- pnpm lint:i18n`, and `mise exec node@22 -- pnpm check:public-boundary` — passed.
- `pnpm install --frozen-lockfile` remains blocked by an existing latest-`main` mismatch: `apps/cli/package.json` requires `loro-crdt` 1.15.1 while the lockfile records 1.14.1. Validation used a temporary regenerated install graph and restored the repository lockfile afterward.
- Not run after this review change: local desktop manual regression, mobile project sheet flow, and the complete cross-package suite.

- **Review focus:** Verify that `loro-app-sidebar.tsx` makes every project-row activation fresh and that the route and `chat-landing.tsx` carry the same selection identity.
- **Decisions to challenge:** Assess the URL intent key instead of synchronizing Composer clears back into the URL, and preserving typed draft text instead of using `resetDraftKey`.
- **Plausible failures / evidence gaps:** The full navigation path lacks an end-to-end test, and desktop and mobile interaction were not manually re-run after removing the hover action.

### Authoring context

- **User goal / directives:** Make the existing project-row interaction reliably select its project for the Composer, including after the same project was cleared.
- **Constraints / non-goals:** Preserve typed draft content, authorization gates, project persistence, removal behavior, and mobile behavior.
- **Risk-bearing decisions:** Treat every row click as fresh intent through a URL key while retaining the existing one-way URL preselection model.
- **Destructive or irreversible behavior:** The change only updates client navigation state and does not delete, migrate, or overwrite user data.
- **Deliberately not done or tested:** No Composer-to-URL synchronization or automated end-to-end navigation test was added; desktop and mobile manual regression remain for follow-up verification.
- **Unknowns / confidence:** Focused component tests and static checks are green; residual risk is limited to integration behavior across real router navigation.


packages/components/src/components/loro-app-sidebar.tsx
```diff
@@ -1135,32 +1151,47 @@ export const LocalProjectItem = memo(function LocalProjectItem({
...
+                        type="button"
+                        className={LOCAL_PROJECT_ROW_ACTION_CLASS}
+                        aria-label={newSessionLabel}
+                        onClick={(e) => {
+                          e.preventDefault();
+                          e.stopPropagation();
+                          onStartProjectSession(machineId, project.id);
+                        }}
+                      >
+                        <Plus className="h-3.5 w-3.5" />
```
wibus-wee: Clicking a project row already means selecting that project for the Composer, so the same action should re-apply the selection even when the URL already points to that project. 

Adding a hover-only “New session” button moves expected behavior from the primary row interaction into a secondary, less discoverable control and gives two controls overlapping purposes.

wibus-wee (changes requested):
Thanks for contributing. But I think this might not be the right interaction model.

packages/components/src/components/loro-app-sidebar.tsx
sheepbox8646: Updated in 309599e. The project row is now the only selection interaction and carries a fresh `projectSelection` intent on every activation; the hover-only button and its icon, prop, story, and i18n wiring are removed. Draft text remains intact. I also made same-millisecond intent keys distinct and ran the component typecheck plus all 392 files / 2798 tests.

packages/components/src/components/chat/chat-landing-derived.ts
```diff
@@ -575,3 +575,35 @@ export function getChatLandingVisibleComposerStatus<TMessage>({
...
+   * pre-selection once per key and then lets the user steer freely, so a
+   * repeated click on the project the URL already names needs a new identity.
+   */
+  projectSelectionKey?: string | undefined;
+};
+
+let projectSelectionSequence = 0;
+
+/** A fresh identity for every project-row activation, including same-millisecond clicks. */
+export function createChatLandingProjectSelectionKey(now = Date.now()): string {
```
wibus-wee: Every activation writes a unique `projectSelection` value without replace, so repeated clicks on the same project push visually identical history entries. 

Going Back then visits the previous nonce and re-applies the same project again. 

It is recommended to use `push` only when switching to a different page or project, if the current URL already belongs to the same project, use `replace: true` to update the intent. 

You can use the existing `selectedLocalProjectKey` to determine whether the target is the same.

packages/components/tests/chat-landing-derived.test.ts
wibus-wee, line 1: The current tests only exercise the key helpers, so they would still pass if either side of the actual navigation wiring were removed.

wibus-wee (changes requested):
Thanks. The revised interaction model now makes sense to me. 

I found some remaining implementation issues before approval.

packages/components/src/components/chat/chat-landing-derived.ts
wibus-wee: Furthermore, I feel that `projectSelection=` is essentially a one-off UI event—simply a way to signal to the already-mounted Composer that "the user has clicked again."

It doesn't seem like something that warrants being copied, bookmarked, or displayed in the address bar; using the query string as an event bus in this implementation strikes me as a bit odd.  🤔

I don’t think we need a separate selection intent at all. The repeated-click problem only exists because clearing or changing the desktop Composer selection leaves the old project in the URL.

When the user explicitly clears the project, the URL should stop naming that project. When they select another project, the URL should name the new one. A later project-row click will then produce an ordinary search-param change and the existing preselection effect will apply it naturally.

That removes the need for projectSelection, the nonce generator, the sequence counter, and the additional preselection-key plumbing entirely.

Or, if you have a better idea, feel free to share it with me! :)

wibus-wee (approved):
LGTM! The CI error seems to be a problem with main, and not yours. Thank you for your contribution.

PR #94, opened 2026-08-27
fix(cli): keep removed builtin agent providers removed across restarts

### LINKED ISSUE #98: [Bug] Deleted builtin agent provider (e.g. Kimi Code) is recreated on every restart
### Affected area

CLI / daemon

### Installation method

Built from source

### Lody version or commit

991c3b2

### Operating system

Windows 11 x64

### Agent or runtime

Managed builtin providers: Claude Code, Codex, Grok, Kimi Code

### What happened?

Deleting one of the four managed builtin providers in Settings → Agents only lasts until the next Lody restart. On startup the provider is back in the list.

Root cause: startup auto-registration (`Lody.ensureBuiltinAgentConfigs`) checks "does this machine have a builtin config of type X?". A "no" cannot tell "never created" apart from "the user just deleted it", so it recreates the config.

The same happens when a provider setup that already published its config is cancelled: the config row is hard-deleted and the next start brings it back.

### What did you expect?

A provider I deleted stays deleted across restarts on that machine. Adding it back explicitly (from the list or via provider setup) re-enables normal behavior.

### How can we reproduce it?

1. Start Lody; Settings → Agents shows Claude Code, Codex, Grok, Kimi Code.
2. Delete Kimi Code.
3. Restart Lody.
4. Observe: Kimi Code is listed again.

### How often does it happen?

Every time

### Relevant log output

```shell
(none — no error is logged; the config is silently recreated)
```

### Before submitting

xuxu777xu:
## Related issue

Closes #98

## Problem / pressure

Settings → Agents lists the four managed builtin providers (Claude Code, Codex, Grok, Kimi Code). Deleting one of them (e.g. Kimi Code) only lasts until the next Lody restart: startup auto-registration in `Lody.ensureBuiltinAgentConfigs` asks "does this machine have a builtin config of type X?", and "no" cannot distinguish "never created" from "the user just deleted it", so the provider is recreated every time.

## Summary

- Add a `builtinAgentOptOut` row family to the machine Flock document (`packages/shared/src/machine-flock.ts`): key `['builtinAgentOptOut', <ManagedBuiltinAgentType>]`, value `{ v, agentType, machineId, removedAt }`, with parse/normalize/getter and a key/value mismatch guard.
- Own the opt-out rules in one place in `@lody/shared`: `planBuiltinAgentOptOutForDeletedConfig` / `findBuiltinAgentOptOutToRetract` (pure, row-map based) plus `writeAgentConfigToFlock` / `deleteAgentConfigFromFlock` (row + tombstone in one commit). The tombstone is written only when the last config of that type on the machine goes away, is never rewritten once present, and is retracted whenever a config of that type is written.
- Route every agentConfig write/delete path through those helpers: CLI `upsertMachineAgentConfig` / `deleteMachineAgentConfig`, provider-setup publish (`provider-setup-manager.ts`) and cancellation (`applyProviderSetupCancellationToFlock`), and the renderer atoms in `packages/components/src/atoms/agents.ts` (which only pay the extra writer round trip when the local mirror actually holds a tombstone; the create atom replaces the optimistic cache so a retracted tombstone leaves it).
- Startup registration (`apps/cli/src/lib/lody.ts`) reads the opt-out set via `LoroDocumentManager.getBuiltinAgentOptOuts` and skips those types before the existence check.
- Non-managed configs (custom ACP, registry, DeepSeek) never write an opt-out.

## Before / after

| Before | After |
| ------ | ----- |
| Delete Kimi Code → restart Lody → Kimi Code is back | Delete Kimi Code → restart Lody → it stays deleted |
| Cancel a published provider setup → restart → provider is back | Cancellation records the same opt-out as a list delete |
| n/a | Add Kimi Code again (list or provider setup) → opt-out cleared; future restarts behave as before |

## Test plan

- `pnpm --filter @lody/shared exec vitest run tests/machine-flock.test.ts` — 25 passed (10 new: round-trip, key/value mismatch rejected, non-managed type rejected, last-of-type delete writes tombstone in one commit, one-of-several does not, repeated delete does not rewrite, write retracts in one commit, identical rewrite is a no-op, cancelling a published setup writes tombstone, cancelling an unpublished one does not)
- `pnpm --filter lody exec vitest run tests/lody-register-agent.test.ts tests/agent-config-machine-flock.test.ts src/lib/provider-setup-manager.test.ts` — 18 passed
- `pnpm --filter @lody/components exec vitest run tests/agent-config-create.test.ts tests/agent-config-dialog.test.tsx` — 16 passed
- `tsgo --noEmit` for `@lody/shared`, `lody`, `@lody/components` — pass
- `oxlint --type-aware` on touched files — 0 errors, warning count unchanged from `main`
- Prettier on every touched file — pass
- Skipped: full `pnpm test:ci` on Windows shows 7 pre-existing failures in unrelated files (`local-cli-host-lease`, `lody-mcp-server`, `acp-authentication`, `message-handler-image-upload`); verified identical failures on `main` without this change. Not verified end-to-end in the Electron app.

- **Review focus:** `packages/shared/src/machine-flock.ts` `planBuiltinAgentOptOutForDeletedConfig` / `deleteAgentConfigFromFlock` (the only place that decides when a tombstone exists) and the four call sites that now go through it: CLI `agent-config-machine-flock.ts`, `provider-setup-manager.ts` publish, `applyProviderSetupCancellationToFlock`, and the renderer atoms in `packages/components/src/atoms/agents.ts`.
- **Decisions to challenge:** Opt-out is per machine and per agent type, written only when the last config of that type is removed; the renderer create atom switched the optimistic cache from `merge` to `replace` so a retracted tombstone actually leaves the cache.
- **Plausible failures / evidence gaps:** Renderer writes go through `runtime.writer` as separate commits rather than one transaction; the renderer decides "tombstone present?" from the local mirror, which can lag in cloud mode; no end-to-end run in the Electron app.

### Authoring context

- **User goal / directives:** Fix "deleted Kimi Code comes back after restart"; then review the PR and fix every path that could still recreate or leave a stale opt-out, keeping the rules in one shared place.
- **Constraints / non-goals:** No fallback/compat branches; no change to which providers are managed builtins; no per-config-id tombstones.
- **Risk-bearing decisions:** Introduces a new persisted row family in the machine Flock document; older clients ignore unknown rows via `parseMachineFlockRow`. Provider-setup cancellation of a published config now also records an opt-out.
- **Destructive or irreversible behavior:** None beyond the existing hard delete of the config row; the opt-out row is removed automatically when the user adds the type back.
- **Deliberately not done or tested:** Full `test:ci` on Windows (pre-existing unrelated failures); manual Electron verification.
- **Unknowns / confidence:** High confidence in the logic and unit coverage; medium on cloud-mode local-mirror lag in the renderer paths.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,4 +154,3 @@ apps/electron/resources/cli/
 /.claude/skills/apple-appstore-reviewer
 /.claude/skills/appstore-info-generator
 /skills-lock.json
-.ace-tool/
--- a/.gitignore
+++ b/.gitignore
@@ -154,3 +154,4 @@ apps/electron/resources/cli/
 /.claude/skills/apple-appstore-reviewer
 /.claude/skills/appstore-info-generator
 /skills-lock.json
+.ace-tool/
--- a/apps/cli/src/lib/agent-config-machine-flock.ts
+++ b/apps/cli/src/lib/agent-config-machine-flock.ts
@@ -1,5 +1,5 @@
 import {
-  deleteMachineFlockRowFromFlock,
+  deleteAgentConfigFromFlock,
   getAgentConfigRoomId,
   getMachineFlockAgentConfigs,
   getMachineFlockBuiltinAgentOptOuts,
@@ -8,7 +8,6 @@ import {
   isAgentConfigDocRoomId,
   isLoroRepoDocDeleted,
   isMachineDocRoomId,
-  isManagedBuiltinAgentType,
   MACHINE_DOC_PREFIX,
   machineFlockKeys,
   readMachineFlockRowsFromFlock,
@@ -18,24 +17,13 @@ import {
   type ManagedBuiltinAgentType,
   type MachineId,
   type WorkspaceId,
-  writeMachineFlockRowToFlock,
+  writeAgentConfigToFlock,
 } from '@lody/shared';
 import type { LoroRepo } from 'loro-repo';
 
 export type MachineFlockSyncScheduler = {
   markMachineFlockDocDirty: (machineId: MachineId, options?: { reason?: string }) => void;
 };
-
-/**
- * 配置指向的托管内置 provider 类型；自定义、registry、DeepSeek 等不参与启动自动注册
- * 的一律返回 null。
- */
-function getManagedBuiltinOptOutAgentType(config: AgentConfigMeta): ManagedBuiltinAgentType | null {
-  if (config.cliType !== 'builtin' || !isManagedBuiltinAgentType(config.agentType)) {
-    return null;
-  }
-  return config.agentType;
-}
 
 /** 本机被用户主动删掉、不该在启动时自动补回来的托管内置 provider 类型。 */
 export async function readMachineBuiltinAgentOptOuts(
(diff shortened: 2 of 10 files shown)


Leeeon233:
@xuxu777xu  We prefer to use English for comments as well.

xuxu777xu:
Thanks — switched every comment on this branch to English in 853ca36.

That commit also fixes the failing `provider-setup-writer` test: planning the builtin agent opt-out needed a mirror read, and I had placed it before the durable cancellation marker write, which pushed the marker past an await. The marker is now the first write again, and the opt-out is planned from the mirror afterwards.

Leeeon233:
I think we can remove agentType, machineId from the value, bacause they can be infered from the key.

xuxu777xu:
Completed in 4c91b8c.

`BuiltinAgentOptOut` now contains only `{ v, removedAt }`: the provider type serves as the second part of the key, while the machine itself constitutes the flock document. The value/key consistency check in `parseMachineFlockRow` has been removed as there is no longer anything to compare against; relevant tests now only check the single failure scenario—rejecting keys that specify a type outside the scope of "auto-register on startup."

Tests for `@lody/shared` and type checks for `shared/cli/components` have passed.

PR #120, opened 2026-08-28
fix(components): keep Codex implementation below plan

ImSingee:
## Related issue

Closes #114

This PR addresses only item 2: implementation progress rendering above the Proposed Plan. The Plan-mode exit failures in items 1 and 3 are intentionally separate. At the time of this update, the issue contains no maintainer comment explicitly approving this scope or approach; this PR does not claim otherwise.

## Problem / pressure

Codex keeps planning and implementation in one assistant turn, but represents the plan as a standalone `proposed_plan` item immediately before the `switch_mode` approval:

```text
planning work -> proposed_plan -> switch_mode -> implementation work -> final answer
```

`buildAssistantMessageRenderItems` previously applied a stable "plans last" partition to the entire turn. That moved the standalone plan past every later item, producing:

```text
planning work -> switch_mode -> implementation work -> final answer -> proposed_plan
```

PR #31 correctly split folding regions at `switch_mode`, but its fixture matched Claude's representation, where the plan is carried by the switch card itself. Because that shape has no standalone `proposed_plan`, it did not expose the Codex-specific global reorder.

## Summary

- Split render ordering into regions delimited by ACP tool calls whose `kind` is `switch_mode`.
- Keep the existing "plan last" presentation only within each region, then leave the switch card and all following implementation work in their original region.
- Preserve each item's source `itemIndex`; only `displayIndex` is recomputed for the rendered order.
- Cover finished and streaming Codex turns, repeated plan/review cycles, and the existing Claude representation.
- Record the segment-local ordering contract in the scoped renderer maintainer guide.

This is a renderer-only change. It does not modify ACP events, persisted session documents, Plan-mode selection, or turn segmentation.

## Before / after

| Before | After |
| ------ | ----- |
| The global partition moved Proposed Plan below implementation progress.<br><br>Implementation progress rendered above Proposed Plan | Proposed Plan remains above its approval card and implementation region.<br><br>Proposed Plan rendered before implementation progress |

## Test plan

- `pnpm --filter @lody/components typecheck` — passed on rebased HEAD.
- `pnpm --filter @lody/components exec vitest run tests/assistant-message-render-items.test.ts tests/assistant-turn-render-blocks.test.ts tests/plan-turn-virtual-rows.test.ts tests/message-copy.test.ts` — 4 files and 42 tests passed.
- `git diff --check origin/main...HEAD` — passed.
[...]

PR #123, opened 2026-08-28
fix: exit Codex plan mode for execution actions

### LINKED ISSUE #114: [Bug] Codex - "implement" a plan doesn't quit plan mode
### Affected area

Desktop app

### Installation method

Desktop release

### Lody version or commit

0.87.0

### Operating system

macOS

### Agent or runtime

_No response_

### What happened?

https://discord.com/channels/1439990585309003798/1537421567141744650

1. Click "Implement this plan" doesn't quit plan mode
2. The implement progress shows above proposed plan
3. Also, click something like "Create PR" button also doesn't quit plan mode

### What did you expect?

Do these things should quit plan mode automatically.

### How can we reproduce it?

1. Start a chat with codex + plan
2. Tell it to make a plan
3. Click "Implement this plan"

### How often does it happen?

Every time

### Relevant log output

```shell

```

### Additional context

#31 fixed the problem for Claude, but not for Codex

### Before submitting

ImSingee:
## Related issue

Closes #114

This PR addresses items 1 and 3: approving a Codex plan and choosing an explicit execution action should both leave Plan mode. Item 2, the Proposed Plan render order, is handled separately by https://github.com/LodyAI/Lody/pull/120.

The author reports that a project maintainer invited opening this PR in private project chat. Issue #114 itself currently contains no public maintainer approval comment, and this PR does not claim otherwise.

## Problem / pressure

Plan is a per-turn run configuration, while approving `switch_mode` changes only the ACP turn that is already running. The composer's next-turn selection is reconstructed from the latest user turn, which still says Plan, so the next message can silently plan again.

The existing frontend exit path only recognized mode-based agents whose `selectedModeId` was the ACP Plan permission mode. Codex represents the same choice as `configOptionValues.collaboration_mode=plan`, so approving "Yes, implement this plan" did not update the Codex selector.

Explicit execution actions had the same mismatch. Implement Plan, Create PR, Create Draft PR, Commit & Push, Resolve Conflicts, Fix CI, and Fix a review finding all reused generic prompt dispatch and therefore preserved the current Plan selection even though their intent is execution.

There was also a replacement-session path back into stale configuration: an adapter can change its own live config and publish a full `config_option_update`, but `AgentClient` previously forwarded that update without refreshing the config retained for a later replacement session.

## Summary

- Consume successful plan-exit approvals in the composer instance that owns run-config state, after its selection has reconciled.
- Handle both representations: Codex exits to `collaboration_mode=default`, while mode-based agents resolve their default non-Plan mode.
- Keep the approval pending until a non-Plan destination exists and consume only the revision that was observed.
- Route explicit Codex execution actions through a Plan-disabled dispatch override and retain the local non-Plan selection only after dispatch is accepted.
- Leave manual sends, Continue discussing, and generic quick messages unchanged so an explicit user Plan choice is not disabled globally.
- Treat ACP `config_option_update` as an authoritative full snapshot in `AgentClient`, rebuilding retained values so removed options cannot leak into replacement sessions.

No shared schema or ACP protocol changes are introduced.

## Before / after

| Before | After |
| ------ | ----- |
| Approving a plan changed only the running ACP turn; the Codex composer still showed Plan and the next turn could plan again. | The approval updates the local Codex collaboration setting to Default once reconciliation is ready. |
| Create PR and other execution actions inherited `collaboration_mode=plan`. | Explicit execution actions dispatch with `collaboration_mode=default` and retain it only after acceptance. |
| Replacement sessions could reuse stale or removed config values after an agent-originated update. | Full agent config snapshots replace the retained option map used for replacement startup. |

## Test plan

- `pnpm --filter lody typecheck` — passed on the rebased head.
- `pnpm --filter @lody/components typecheck` — passed on the rebased head.
- `pnpm --filter @lody/components exec vitest run tests/codex-plan-decision.test.ts tests/plan-mode-exit.test.ts` — 2 files and 14 tests passed.
- `pnpm --filter lody exec vitest run src/agent/agent-client-session-preparation.test.ts` — 1 file and 10 tests passed.
- `git diff --check origin/main...HEAD` — passed.
- The correctly rebased Plan-exit branch was not manually clicked through in the desktop app; verification is type-level and at the dispatch/config-state helper boundaries.

- **Review focus:** Trace approval consumption in `use-plan-mode-exit-approval.ts`, execution dispatch wiring in `session-chat-interface.tsx`, and authoritative replacement-session state in `agent-client.ts`.
- **Decisions to challenge:** Validate the click-local signal, explicit execution-action allowlist, accepted-dispatch commit point, and full-snapshot replacement semantics.
- **Plausible failures / evidence gaps:** Composer remount and concurrent run-config edits remain known gaps, and the rebased branch has no final desktop click-through.

### Authoring context

- **User goal / directives:** Fix Codex Plan-mode exit for plan approval and execution actions in a PR separate from the Proposed Plan ordering change.
- **Constraints / non-goals:** Preserve Plan for manual discussion sends, avoid shared schema or protocol changes, and do not combine PR #120's renderer fix.
- **Risk-bearing decisions:** Approval state is tab-local to avoid history sync changing a teammate's selector; explicit execution dispatch overrides Plan only for accepted turns; agent config updates replace retained state as authoritative snapshots.
- **Destructive or irreversible behavior:** The change performs no migration, deletion, external write beyond ordinary prompt dispatch, or irreversible operation; reverting restores the former selection behavior.
- **Deliberately not done or tested:** Per prior scope direction, this PR does not persist the post-exit preference across a composer remount and does not merge only the Plan key when an async execution action resolves; no final desktop interaction test was run.
- **Unknowns / confidence:** Confidence is high for the immediate approval, explicit-action, and replacement-session paths; navigating away before a new turn can restore Plan, and a delayed action can overwrite newer Reasoning or Fast changes with its captured config map.

### DIFF
--- a/apps/cli/src/agent/agent-client-session-preparation.test.ts
+++ b/apps/cli/src/agent/agent-client-session-preparation.test.ts
@@ -1,5 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import type { MachineId, SessionId, WorkspaceId } from '@lody/shared';
+import type { SessionNotification } from '@agentclientprotocol/sdk';
 import type { Logger } from '@/utils/logger';
 
 const connectionMocks = vi.hoisted(() => ({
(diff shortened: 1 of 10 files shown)


wibus-wee (commented):
Thanks for your contribution! I have a few thoughts on the PR and the issue. :)

_"Yes, implement this plan"_ resolves a permission within the existing turn, so the latest durable user turn still records Plan.

The PR updates the mounted Composer to Default, but consumes the local approval before another user turn makes that choice durable. As a result, remounting the Composer first can restore Plan.

After reviewing your issue and PR, it seems the intended behavior is for the switch to Default to remain effective until the user explicitly selects Plan again.

So I think this is a non-blocking comment: we should retain a tab-local next-turn override across remounts and clear it only after a Default user turn is accepted, or when the user explicitly selects Plan again.

ImSingee:
@wibus-wee Thanks for your review! I've fixed this.

wibus-wee:
Hi @ImSingee, thank you again for working through this with us.

Your issue and PR made the Codex Plan-exit gap concrete, and the investigation here helped us identify that the problem goes beyond the mounted composer.

While validating remounts and multi-client collaboration, we realized that the Plan exit should not be inferred from the UI action itself. The authoritative transition is the configuration confirmed by the ACP agent, and that confirmed runtime state needs to be projected into the shared Session document so every collaborator sees the same baseline without overwriting local unsent choices.

I opened #159 with that shared-state approach. It directly builds on the problem analysis and implementation work from this PR, so I have also added you as a co-author there.

I would especially appreciate your thoughts on the ACP notification and replacement-session paths. 

Thanks for pushing this issue forward. ❤️

PR #143, opened 2026-08-28
fix(desktop): restore input focus after sending message

### LINKED ISSUE #131: [Bug] The input lost focus after sending message
### Affected area

Desktop app

### Installation method

Desktop release

### Lody version or commit

0.87.0

### Operating system

macOS

### Agent or runtime

_No response_

### What happened?

After sending a message, the input box will lose focus.

### What did you expect?

After sending the message, focus remains on the input field so you can continue typing.

### How can we reproduce it?

1. Send message
2. Observed that focus was lost

### How often does it happen?

Every time

### Relevant log output

```shell

```

### Additional context

_No response_

### Before submitting

clanzhang:
## Related issue

Closes #131

## Problem / pressure

On macOS Desktop (v0.87.0), after sending a message in a session, the input textarea loses focus. Users must manually click the input area to continue typing, breaking the conversation flow. 100% reproducible.

## Summary

Added focus restoration to the chat input textarea after message send completes on desktop platforms. Mobile behavior is unchanged — the keyboard still dismisses on send and focus is only restored when the send is rejected.

## Before / after

| Before | After |
| ------ | ----- |
| Input loses focus after sending; user must click to continue typing | Input retains focus after sending; user can immediately type the next message |

## Test plan

- [x] macOS Desktop: Send message via Enter key → input keeps focus
- [x] macOS Desktop: Send message via button click → input keeps focus
- [ ] iOS/Android: Send message → keyboard dismisses (unchanged)
- [ ] iOS/Android: Failed send → focus restored with draft preserved

- **Review focus:** `packages/components/src/components/sessions/session-chat-input-area.tsx` — the `finally` block in `sendMessage` (around line 1820) where focus restoration was added
- **Decisions to challenge:** Whether `!dismissKeyboardForSubmit` is the right condition to distinguish desktop from mobile, or if an explicit platform check would be clearer
- **Plausible failures / evidence gaps:** No automated test added for desktop focus restoration; relies on manual verification. Existing mobile test at line 270 of `session-chat-input-submission.test.tsx` should still pass.

### Authoring context

- **User goal / directives:** Fix issue #131 where desktop input loses focus after send
- **Constraints / non-goals:** Must not change mobile keyboard behavior. Must not affect message send logic or draft persistence.
- **Risk-bearing decisions:** Focus restoration in `finally` block ensures it runs on both success and failure paths. Uses existing `dismissKeyboardForSubmit` flag to detect mobile vs desktop.
- **Destructive or irreversible behavior:** None — only adds a focus call, no state mutations.
- **Deliberately not done or tested:** No automated test for desktop focus (jsdom focus behavior may not match real browser). Mobile rejected-send focus restoration already covered by existing test.
- **Unknowns / confidence:** High confidence for desktop fix. Mobile behavior unchanged and covered by existing test. Edge case: if textarea unmounts during send, optional chaining `?.focus()` safely no-ops.

### DIFF
--- a/packages/components/src/components/sessions/session-chat-input-area.tsx
+++ b/packages/components/src/components/sessions/session-chat-input-area.tsx
@@ -518,6 +518,10 @@ export const SessionChatInputArea = memo(
     const isArchived = session.isArchived === true;
     const textareaRef = useRef<HTMLTextAreaElement>(null);
     const restoreFocusAfterRejectedMobileSendRef = useRef(false);
+    /** Session id that initiated the pending desktop focus restore. */
+    const pendingDesktopFocusRestoreSessionIdRef = useRef<string | null>(null);
+    /** The textarea element that was active when the send started. */
+    const pendingFocusRestoreTextareaRef = useRef<HTMLTextAreaElement | null>(null);
     const attachmentInputRef = useRef<HTMLInputElement>(null);
     const activeSessionIdRef = useRef(session.id);
     activeSessionIdRef.current = session.id;
@@ -1791,6 +1795,12 @@ export const SessionChatInputArea = memo(
         ];
         const dismissKeyboardForSubmit =
           usesMobileKeyboardAction && (source === 'keyboard' || source === 'button');
+        // Snapshot the textarea and session id BEFORE the await so the
+        // post-commit focus-restore effect can verify neither changed during
+        // the in-flight send. Capturing after the await would miss a session
+        // switch that happened while the send was pending.
+        const textareaBeforeSend = textareaRef.current;
+        const sessionIdBeforeSend = session.id;
         if (dismissKeyboardForSubmit) {
           // The mobile Send action should dismiss the soft keyboard at the same
           // immediate handoff boundary as the visible draft, not after the
@@ -1819,7 +1829,16 @@ export const SessionChatInputArea = memo(
           }
         } finally {
           restoreFocusAfterRejectedMobileSendRef.current = dismissKeyboardForSubmit && !accepted;
+          if (!dismissKeyboardForSubmit) {
+            // Stash the pre-send snapshot for the focus-restore effect below.
+            pendingDesktopFocusRestoreSessionIdRef.current = sessionIdBeforeSend;
+            pendingFocusRestoreTextareaRef.current = textareaBeforeSend;
+          }
           setSubmissionPending(false);
+          // Focus is NOT restored synchronously here: the textarea is still
+          // disabled (React has not yet committed the re-render that clears
+          // submissionPending). The desktop focus-restore useEffect below
+          // handles it after the re-enable render.
         }
       },
       [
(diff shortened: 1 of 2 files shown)


Leeeon233:
Thanks for working on this. There is a timing issue in the current implementation that means it does not reliably restore desktop focus.

`setSubmissionPending(false)` only schedules the React update. The following `textareaRef.current?.focus()` therefore runs while the textarea is still rendered with `promptDisabled={submissionPending || isArchived}`. A disabled form control cannot be focused, so the call is ignored; the next render enables the textarea, but nothing focuses it afterward.

There is also a stale-completion case: this composer can switch `session.id` in place while a send is pending. When the old promise settles, `textareaRef.current` may already point to the new session textarea, causing the old submission to steal focus in a different session or override another control the user deliberately focused.

Could you please update the fix to:

1. Record a desktop focus-restore intent in `finally`, then perform the focus from an effect/layout effect only after `submissionPending` is false and the textarea is enabled. The existing rejected-mobile-send effect below this block demonstrates the required post-commit timing.
2. Bind the restore intent to the session that submitted the message, clear or ignore it after a session change, and avoid restoring focus if the user deliberately moved focus during the wait.
3. Add regression coverage for desktop Enter and button submission with a deferred `onSendMessage`, plus switching sessions or focusing another control before the promise resolves. Please also preserve the existing mobile behavior: successful sends stay blurred, while rejected sends restore focus.

clanzhang:
Thanks for the detailed review — both issues are valid.

**Timing issue**: You're right. `setSubmissionPending(false)`  only schedules the update, so the synchronous `focus()`  in the same tick ran while the textarea was still `disabled` . I've moved the focus into a post-commit `useEffect`  that fires after `submissionPending`  flips to `false` , so the textarea is re-enabled by the time `focus()`  runs.

**Stale session**: Good catch. I've added a `pendingFocusRestoreTextareaRef` that captures the textarea DOM element at send time. The restore effect compares `textareaRef.current` against that captured reference — if the component unmounted and remounted with a different session, the refs no longer match and the restore is skipped.

**User focus changes**: The restore effect also checks `document.activeElement !== capturedTextarea`  before focusing, so if the user deliberately moved focus to another control during the wait, the restore is skipped.

**Mobile behavior preserved**: The existing `restoreFocusAfterRejectedMobileSendRef`  path is untouched — successful mobile sends still stay blurred, rejected sends still restore focus.

**Regression tests added** :

- Desktop Enter send restores focus
- Desktop button click restores focus
- User moves focus during pending send → not restored
- Session changes (unmount/remount) during pending send → not restored

Leeeon233:
Thanks for the update. Two correctness issues remain:

- Please remove the synchronous `focus()` in `finally`. The textarea is still disabled there. The post-commit effect also should not treat `document.activeElement !== textarea` as proof that the user moved focus, because disabling the textarea normally moves focus to `body`. Restore focus after re-enable when the same session is still active and no other interactive control owns focus.
- The session guard is not currently effective: the stored session id is never compared, and the textarea is captured after `await`. Capture the restore target before the send, compare the stored id with the current session, and test an in-place session prop change after actually starting a pending send. The current session-switch test never triggers a send.

Please also fix the failing lint/static checks before requesting another review. `originalTextarea` is unused, the session-switch test contains duplicated/unformatted code, and the diff has trailing whitespace. Please run the repository formatter and `pnpm check:quick` (or the full `pnpm check`) and push a clean result.

wong2:
Has this been released? I still have this issue in version 0.91.1.

Leeeon233:
> Has this been released? I still have this issue in version 0.91.1.

Let me check it.  There still seem to be some race conditions.

PR #159, opened 2026-08-29
fix: synchronize ACP runtime config across collaborators

wibus-wee:
## Related issue

https://github.com/LodyAI/Lody/issues/114

This is the shared-state alternative to the click-local approach in https://github.com/LodyAI/Lody/pull/123.

## Problem / pressure

Approving "Yes, implement this plan" changes the running ACP agent from Plan to Default, but the next-turn composer was reconstructed from the latest user turn and could remain in Plan. A local component signal cannot survive remounts or keep collaborators consistent, and it can incorrectly overwrite unsent settings such as a teammate's Reasoning choice.

Shared runtime state alone also cannot express the intent of a new execution action before that Turn is accepted. Create PR, Commit & Push, Fix CI, and similar actions must therefore freeze a supported non-Plan mode into their own Turn input.

## Summary

- Treat ACP config responses plus `config_option_update` and legacy `current_mode_update` notifications as the runtime source of truth.
- Project confirmed runtime changes into a revisioned `SessionDoc.acpRuntimeConfig` snapshot fenced to the driving user turn.
- Filter sensitive option IDs and unsupported option types before shared persistence, with a second guard at the `SessionDocument` durable boundary.
- Treat every present `configOptions` response, including `[]`, as an authoritative full snapshot. When a legacy agent accepts a request but omits that field, apply the accepted value to both replacement startup state and the matching advertised option used by runtime projection.
- Reconcile the shared baseline into every composer while preserving local unsent edits and keeping accepted or queued Turn input immutable.
- Carry agent-originated config changes into replacement ACP sessions and persist request-confirmed state after run-config application.
- Freeze a supported non-Plan mode into explicit execution Turns without mutating or resetting the local composer draft.

## Before / after

| Before | After |
| ------ | ----- |
| Plan exit changed only one mounted composer and could be lost on remount. | Agent-confirmed Default is stored in the shared session document and survives remount. |
| One collaborator's local update could not inform another client. | Independent Loro replicas converge on the same runtime baseline. |
| Applying shared state risked replacing a local unsent Reasoning choice. | Shared fields reconcile independently; local unsent edits remain private until send. |
| Execution actions could create their next Turn in Plan before runtime state changed. | Implement Plan and PR/CI execution actions freeze a supported non-Plan mode into that Turn while preserving Reasoning and other option values. |
| A future ACP string option could persist a credential-like value in shared state. | Sensitive IDs and unsupported option shapes are removed both during projection and at durable persistence. |
| An explicit empty ACP snapshot retained stale values for replacement sessions. | `configOptions: []` clears retained values; a legacy omitted snapshot keeps the accepted value consistent across shared projection and replacement startup. |

## Test plan

- 86 focused Vitest assertions passed: 42 CLI, 21 components, and 23 shared.
- The full CLI suite passed: 242 test files, 2,402 assertions passed, and 4 skipped.
- The two-client integration test uses independent `LoroDoc` and `Mirror` instances with real export/import updates; it verifies Plan to Default convergence and proves one client's unsent High reasoning does not leak to the other client.
- Execution-Turn tests verify Codex config-option Plan, legacy mode Plan, advertised fallback selection, unsupported no-fallback behavior, and preservation of Reasoning/Fast values.
- Runtime regression tests verify credential-like IDs never reach the shared document, an authoritative empty set-config snapshot cannot reappear in a replacement ACP session, and a legacy omitted snapshot publishes and restarts with the same accepted value.
- A temporary, uncommitted real Codex ACP harness passed: start Plan + Low, accept the actual `implement_plan` action, observe the agent's Default + Low config update, and verify the requested implementation side effect.
- CLI and components typechecks passed.
- `pnpm --filter lody format:check`, targeted Prettier checks, `git diff --check`, and targeted Oxlint completed with 0 errors. Public, platform, and Code Collab boundary guards passed.
- The root `pnpm check` wrapper could not start locally because this environment has no `corepack`; its relevant checks were invoked directly.

- **Review focus:** Trace ACP response/notification projection through `runtime-config.ts`, `message-handler.ts`, `SessionDocument`, and composer reconciliation. Separately trace explicit execution actions through `dispatchPrompt` into the accepted Turn's `inputConfig`.
- **Decisions to challenge:** Validate that accepted/queued Turns remain immutable, local unsent fields win per key, full snapshots remove omitted non-user options without erasing explicit drafts, legacy acknowledgements keep runtime and replacement state aligned, and sensitive values cannot cross the durable shared boundary.
- **Plausible failures / evidence gaps:** There is no committed live two-Electron-window transport test. An agent that returns no response state and emits neither a standard nor legacy config notification cannot publish a runtime transition. An agent that advertises Plan as its only mode has no supported execution destination, so the client deliberately does not invent one.

### Authoring context

- **User goal / directives:** Fix Codex Plan exit as shared collaborative state rather than a local Implement-button side effect, while ensuring explicit execution actions cannot accidentally create a Plan Turn.
- **Constraints / non-goals:** Keep ACP compatibility, preserve local unsent choices, avoid changing frozen Turn input after acceptance, do not invent unsupported agent modes, and do not include PR #120's Proposed Plan rendering work.
- **Risk-bearing decisions:** Agent-confirmed runtime state is a revisioned shared session snapshot causally tied to the latest user Turn. The intent to execute is frozen separately into the newly dispatched Turn, because shared post-confirmation state cannot precede that action.
- **Destructive or irreversible behavior:** There is no migration or deletion of user data; rollback is reverting these commits, after which readers ignore the additive session field.
- **Deliberately not done or tested:** No permanent network/account E2E was committed because repository tests must be deterministic, and no manual two-window Electron transport run was performed; real ACP and deterministic two-replica boundaries were tested separately.
- **Unknowns / confidence:** Confidence is high for Codex and agents using current or legacy ACP config events. Agents that silently mutate config without returning or notifying it remain unobservable by design.


wibus-wee:
CI status note: both `Static checks` and `Tests` stopped in `pnpm install --frozen-lockfile` before any PR code ran. The current `main` manifest requires `apps/cli` `loro-crdt` 1.15.1 while `pnpm-lock.yaml` still records 1.14.1.

I reproduced the same frozen-install failure locally. A full lockfile regeneration changes 457 lines across unrelated catalog and peer resolutions, so I have not folded that repository-wide baseline repair into this ACP-focused PR. The direct typechecks and 71 focused tests reported in the PR body were run with dependencies resolved from the current manifests.

PR #161, opened 2026-08-29
fix: reapply sidebar project selection

wibus-wee:
## Summary

- make every desktop sidebar project-row activation reapply that project to the mounted Composer
- deliver repeated same-project activations through a transient in-process request instead of URL or history state
- keep the existing stable route search only for navigation, refresh, and deep-link restoration
- leave mobile behavior unchanged and avoid adding a secondary hover-only action

This is an alternative to #87. The project row remains the primary interaction, while Composer-internal selection changes no longer rewrite the URL.

## Testing

- `vitest run packages/components/tests/chat-landing-project-selection.test.ts packages/components/tests/chat-landing-derived.test.ts` (59 tests passed)

Full package typecheck was not run from this worktree because its dependencies are not installed; CI should perform the authoritative check.

### DIFF
--- a/packages/components/src/components/chat/chat-landing-project-selection.ts
+++ b/packages/components/src/components/chat/chat-landing-project-selection.ts
@@ -1,0 +1,35 @@
+export type ChatLandingProjectSelection =
+  | { kind: 'none' }
+  | { kind: 'github'; repoFullName: string }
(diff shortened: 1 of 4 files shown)


wibus-wee:
Superseded by https://github.com/LodyAI/Lody/pull/162. GitHub automatically closed this PR when its head branch was renamed through the branch rename API.

PR #211, opened 2026-08-31
fix: keep title generation on automatic model

### LINKED ISSUE #146: [Bug] Amp title generation defaults to Ultra
### Affected area

Agent runtime / ACP

### Installation method

Built from source

### Lody version or commit

`b9dcace` (the behavior also affects earlier builds containing positional title-model selection)

### Operating system

macOS arm64

### Agent or runtime

Amp ACP

### What happened?

Lody creates an isolated ACP session to generate a conversation title. When no title model is configured, it treats the last advertised model option as the preferred default.

Amp advertises `low`, `medium`, `high`, and `ultra` through a model-category option, with `ultra` last. As a result, an Amp conversation running in Medium can silently start a separate Ultra session only to generate its title. Ultra is Amp's most expensive mode, so using it as an implicit housekeeping default is unnecessarily costly and surprising.

### What did you expect?

Unconfigured title generation should keep the ACP provider's current/default model. ACP option order does not define preference, quality, or cost, so Lody should not infer a model choice from array position.

### How can we reproduce it?

1. Configure Amp ACP without an explicit title-generation model override.
2. Start a conversation in Medium mode.
3. Inspect the ACP sessions in the local logs.
4. Observe that the main conversation uses Medium while the isolated title-generation session is switched to Ultra.

### How often does it happen?

Every time

### Additional context

Proposed direction:

- represent title-generation settings as sparse explicit overrides;
- show Automatic in settings without persisting an inferred model;
- leave the model provider-owned unless the user explicitly selects one;
- lower only recognized permission and reasoning values for the isolated session;
- preserve unknown or legacy values instead of guessing from provider-specific ordering.

A local, unpushed proof of concept was prepared while diagnosing the bug. No external pull request will be opened until a Lody maintainer explicitly agrees on the scope and approach here.

### Before submitting

AmAzing129:
## Related issue

Closes #146

## Problem / pressure

Lody's isolated title-generation session treated the last advertised model option as the preferred default when no title model was configured. ACP option order carries no cost or preference semantics, so an Amp conversation running in Medium silently spawned a separate Ultra session just to generate its title — Amp's most expensive mode as an implicit housekeeping default.

## Summary

Title-generation config becomes a sparse set of explicit overrides; missing keys stay provider-owned ("Automatic").

- `computeTitleGenerationDefaults` no longer selects a model at all: an unconfigured model keeps the ACP session's current/default value. Only recognized `mode` (permission) and `reasoning_effort`/`thought_level` values are lowered, via exact alias matching instead of substring matching; unrecognized vocabularies are left untouched.
- New `applyResolvedTitleConfigOptions` applies an explicit model first, then recomputes automatic overrides from the updated option vocabulary, so a stale reasoning override from a previous model cannot leak into the new model's session. Unavailable persisted overrides stay stored for UI compatibility but never suppress a recognized safe runtime default, and are skipped at apply time.
- Settings UI shows "Automatic (current value)" without persisting an inferred value; selecting Automatic removes the key. Unavailable legacy selections render as "Unavailable (value)" and an unrelated Save no longer rewrites them. The preset backfill (`buildPresetTitleGeneration`) is removed — the runtime treats an absent `titleGeneration` as automatic, and the "not configured" gate the backfill guarded against no longer exists.

## Before / after

| Before | After |
| ------ | ----- |
| Unconfigured title model = last advertised option (Ultra for Amp) | Unconfigured title model keeps the provider's current/default model |
| Substring rank matching could misclassify option names | Exact alias matching; unrecognized values are never overridden |
| Unavailable persisted override suppressed the safe runtime default for that option | Recognized safe default wins; the legacy value stays stored but is not applied |
| Settings persisted inferred defaults on save | Automatic is display-only; only explicit user choices persist |

## Test plan

- `apps/cli`: `vitest run src/agent/title-generator.test.ts` — 13 passed (new coverage: stale explicit reasoning after model switch, sparse merge, unavailable-override handling)
- `packages/shared`: `vitest run tests/title-generation-defaults.test.ts` — 13 passed
- `packages/components`: `vitest run tests/agent-config-dialog.test.tsx` — 18 passed (Automatic display, no persistence of inferred values, unavailable legacy selection preserved)
- `pnpm check:quick` (oxlint, i18n keys, code-collab/platform/public boundary guards) — passed
- `pnpm typecheck` — passed across the workspace
- Not tested: live Amp ACP end-to-end on this exact commit (the reporter's earlier local PoC validated the Medium-session repro from the issue)

- **Review focus:** `packages/shared/src/ai.ts` (`computeTitleGenerationDefaults`, alias tables) and `apps/cli/src/agent/title-generator.ts` (`applyResolvedTitleConfigOptions` ordering: model first, then recomputed defaults from the refreshed vocabulary).
- **Decisions to challenge:** exact alias matching instead of substring matching — nonstandard option names (e.g. "Low effort") are no longer lowered and keep the provider default; removal of `buildPresetTitleGeneration` relies on the runtime treating absent `titleGeneration` as automatic.
- **Plausible failures / evidence gaps:** the legacy `unstable_setSessionModel` path clears the vocabulary and applies no automatic overrides (conservative by design); `getBuiltinTitleGenerationDefaults` is now dead code kept for compatibility, referenced only by tests.

### Authoring context

- **User goal / directives:** stop Ultra from being an implicit title-generation default for Amp; keep unconfigured title generation on the provider's own default, per the scope approved in issue #146.
- **Constraints / non-goals:** no schema migration — existing persisted overrides keep working; no cleanup of the deprecated `getBuiltinTitleGenerationDefaults` in this PR.
- **Risk-bearing decisions:** unavailable persisted overrides no longer suppress recognized safe runtime defaults; presets no longer persist inferred title defaults on save.
- **Destructive or irreversible behavior:** none — stored config is only ever narrowed by explicit user action (selecting Automatic deletes that key).
- **Deliberately not done or tested:** live end-to-end ACP session inspection on this commit; dead-code removal of the compatibility shim.
- **Unknowns / confidence:** high confidence in the core fix and tests; residual risk is limited to providers with nonstandard mode/reasoning vocabularies, which now conservatively keep their own defaults.


Leeeon233:
Hi @AmAzing129 , could you resolve the conflicts?

AmAzing129:
Thanks for the heads-up! AmAzing129 has resolved the conflicts with the latest `main` and pushed the update. All local checks (`pnpm check`) pass.

PR #218, opened 2026-08-31
fix(acp): map pi, grok, and kimi to the skill dirs their engines scan

### LINKED ISSUE #150: [Bug] Skills panel and $ mention miss ~/.agents/skills for pi, grok, and kimi agents (stale per-agent skill-dir whitelist)
### Affected area

Agent runtime / ACP

### Installation method

Desktop release

### Lody version or commit

0.88.0

### Operating system

macOS 26.6.2 arm64

### Agent or runtime

pi via ACP Registry (`npx --prefer-offline -y pi-acp@0.0.33`), grok-build 1.0.0 (Lody-managed binary), kimi-code 0.37.0-lody (managed provider)

### What happened?

The built-in Skills panel (Settings -> Projects -> Skills) and the `$` skill mention picker show **no global skills** for the **pi** (registry `pi-acp`), **grok** (grok-build), and **kimi** (kimi-code) runtimes when skills live in `~/.agents/skills` - the cross-harness Agent Skills standard directory that all three engines discover natively. The agents themselves load the skills correctly; only Lody's UI can't see them.

#### Root cause

`packages/shared/src/acp/skills.ts` maintains a hand-written per-agent whitelist (`ACP_SKILL_DIRS_BY_AGENT_TYPE`). Three entries don't match the engines' actual discovery rules:

1. **pi / pi-acp (lines 157-158)** - `globalDirs` only contains `~/.pi/agent/skills`. pi actually discovers both `~/.pi/agent/skills/` and `~/.agents/skills/` (documented in pi's `docs/skills.md`; also verifiable in a live pi session, where skills from `~/.agents/skills` are present in the system prompt and advertised as `/skill:*` commands).

2. **grok - not registered at all.** grok-build maps to `agentType: 'grok'`, but there is no `grok` key in the table. Unregistered types collapse into a truthy empty whitelist, which filters every scanned skill out of the `$` menu (the same mechanism #76 fixes for custom agents). Grok's own engine scans `~/.grok/skills/` and `~/.agents/skills/` (plus configurable Claude/Cursor compat dirs) - from grok's built-in docs: *"Grok also scans `.agents/skills/` (and `commands/`) at each tier (alongside `.grok/`)"*, with known skill roots `.grok/`, `.agents/`, `.claude/`, `.cursor/`.

3. **kimi (lines 139-141)** - `globalDirs` is an explicit empty array, and `projectDirs` uses `.kimi/skills` while the engine actually uses `.kimi-code/skills`. The kimi-code bundle's skill-scanner constants are: user tier = `<brandHome>/skills` (i.e. `~/.kimi-code/skills`) + `~/.agents/skills`; project tier = `.kimi-code/skills` + `.agents/skills`.

For comparison, ~15 other entries in the same table (codex, gemini, goose, zed, warp, opencode, github-copilot, devin, auggie, mistral-vibe, loaf, ...) already include `DEFAULT_AGENTS_GLOBAL_SKILL_DIR` (`~/.agents/skills`), which is why Codex works out of the box. Claude Code in my setup only works because I personally symlinked `~/.claude/skills -> ~/.agents/skills`.

### What did you expect?

Lody's Skills panel and `$` mention picker should list global skills from `~/.agents/skills` for pi, grok, and kimi sessions, matching what the engines themselves discover and load.

### How can we reproduce it?

1. Install skills into `~/.agents/skills/<name>/SKILL.md`.
2. Start a pi-acp (or grok-build / kimi-code) session in a project.
3. The engine loads the skills (pi's system prompt lists them; grok and kimi discover them as well).
4. Observe that Lody's `$` mention menu / Skills tab shows no global skills for that agent.

### How often does it happen?

Every time

### Relevant log output

```text
2026-08-29T05:52:44.205Z [DEBUG] [start] [322bfc9f-...] Starting ACP agent process (cwd=/Users/.../project cmd=npx args=["--prefer-offline","-y","pi-acp@0.0.33"])
```

The pi-acp startup block lists the skills from `~/.agents/skills`, confirming the engine side loads them - only Lody's UI filtering misses them.

### Additional context

Workaround (what I use today): symlink each engine's native global dir to the shared one:

```bash
ln -s ~/.agents/skills ~/.pi/agent/skills
ln -s ~/.agents/skills ~/.grok/skills
ln -s ~/.agents/skills ~/.kimi-code/skills
```

Suggested fix, two layers:

**P0 (data):** align the three entries with the engines' documented discovery rules:

```ts
pi:       skillDirs(['.agents/skills', '.pi/skills'], ['~/.pi/agent/skills', '~/.agents/skills']),
'pi-acp': skillDirs(['.agents/skills', '.pi/skills'], ['~/.pi/agent/skills', '~/.agents/skills']),
grok:     skillDirs(['.agents/skills', '.grok/skills'], ['~/.grok/skills', '~/.agents/skills']),
kimi / 'kimi-code' / 'kimi-code-cli':
          skillDirs(['.agents/skills', '.kimi-code/skills'], ['~/.kimi-code/skills', '~/.agents/skills']),
```

**P1 (structural, prevents recurrence):** treat the standard directories as an unconditional baseline instead of relying on per-agent completeness - union `.agents/skills` (project) and `~/.agents/skills` (global) into every agent's allowed dirs, keeping the per-agent table only for agent-specific extras. Unregistered agent types (e.g. future registry runtimes) would then fall back to the baseline, which is more precise than the unrestricted `null` proposed in #76, and aligned with the [agentskills spec](https://github.com/agentskills/agentskills) recommendation that every client scan both its native directory and `.agents/skills` at each scope. The existing `KNOWN_SKILL_DIRS_VERSION` cache mechanism (v9 -> v10) can carry this semantic change.

Over-inclusion risk is negligible: all mainstream engines already discover `.agents/skills`, and a `$` mention expands to a `SKILL.md` path in the prompt, which an engine can read even if it doesn't discover the directory itself.

Related: #76 (empty-whitelist filtering for unregistered custom agents - same mechanism, different entry point).

### Before submitting

Astro-Han:
## Related issue

Closes #150

## Problem / pressure

`ACP_SKILL_DIRS_BY_AGENT_TYPE` is a hand-maintained mirror of facts owned by each engine, and it had drifted from what the engines actually scan:

- **pi** (`packages/coding-agent/docs/skills.md`, "Locations") scans `~/.pi/agent/skills` **and** `~/.agents/skills`; the table listed only the first.
- **grok** had no entry at all. `MANAGED_BUILTIN_RUNTIMES` maps `grok-build` to `agentType: 'grok'`, but no `grok` key existed. Its skill roots come from `CompatConfig::skill_config_dirs()` in `xai-grok-tools/src/types/compat.rs`, which returns `[".grok", ".agents", ".claude", ".cursor"]` under the all-on `VendorCompat::default()` (pinned by that file's own `skill_config_dirs_all_on_matches_legacy_constant` test), at every tier.
- **kimi** uses `.kimi-code/skills` + `.agents/skills` per project and `~/.kimi-code/skills` + `~/.agents/skills` globally (`packages/agent-core-v2/src/features/skill/catalog/skillRoots.ts`). The table had `.kimi/skills` and an empty global list. `.kimi/skills` is not in the engine's project tier at all, so it is stale rather than merely incomplete. (The documented first-launch migration copies the *user* dir `~/.kimi/skills`; the project dir was never migrated, it simply stopped being scanned.)
- **deepseek** had no entry either, and it is a `BUILTIN_AGENTS` member. `dsh-skill-filesystem` — mounted by the standard and code presets (`packages/acp-extension-dsh/presets/*/agent.cordis.yml`) with no roots configured — discovers `.dsh/skills` and `.agents/skills` per project and `~/.dsh/skills` and `~/.agents/skills` per user (`skill-filesystem` config defaults: `dshHome` = `$DSH_HOME` or `~/.dsh`, `agentsHome` = `$DSH_AGENTS_HOME` or `~/.agents`).

The grok symptom was worse than a missing row. `getAllowedSkillMentionDirs` returned an **empty but non-null** `Set` for an agent type absent from the table, and `selectSkillMentionCandidates` treats non-null as "apply this whitelist" — so every scanned `$` skill was filtered out. A lookup miss meant "hide everything". The Skills panel, reading the same table via `annotateAndSortGroups`, treats a miss as merely unannotated and hides nothing, so the two surfaces disagreed about the same miss. Custom `custom-<uuid>` agents hit this too.

Nothing tied the table to the authoritative agent lists, so `grok` and `deepseek` could go missing without any check noticing.

## Summary

- Register `grok` and `deepseek`, and align `pi` / `pi-acp` / `kimi` / `kimi-code` / `kimi-code-cli` with their engines' dirs.
- `deepseek` therefore gets the same treatment as the other three: it is a fourth agent that reads `~/.agents/skills` while Lody could not see it. (The `skills: enabled: false` in `profile.ts` disables the demo spine's own bundled wiring — alongside `toolBash`, `goals`, and `persona` — because the host composition mounts the real skill registry and the presets mount the discovery and loader plugins.)
- Return `null` (no whitelist) from `getAllowedSkillMentionDirs` for an unregistered agent type, via a new `isRegisteredSkillAgentType`. A registered entry that maps to nothing keeps its explicit empty whitelist — "reads no skills" and "we have no mapping" stay distinct.
- Add a test asserting every `BUILTIN_AGENTS` type is registered, so a new builtin cannot repeat this.
- Bump `KNOWN_SKILL_DIRS_VERSION` to 10 so cached scans re-fetch.

This revives the `null` change from #76 (closed unmerged under `status:needs-pr-body`, not on technical grounds); #150 names that same mechanism.

## Before / after

| Before | After |
| ------ | ----- |
| `$` menu showed no skills for grok: an unregistered type produced an empty whitelist that filtered every candidate. | An unregistered type applies no whitelist; grok and deepseek are also registered outright. |
| `~/.agents/skills` was invisible to pi, grok, and kimi although all three load it. | All three map to `~/.agents/skills`. |
| kimi mapped to `.kimi/skills` and no global dir. | kimi maps to `.kimi-code/skills`, `~/.kimi-code/skills`, and `~/.agents/skills`. |
| deepseek, a builtin, was absent, so its `$` menu was empty. | deepseek maps to `.dsh/skills`, `.agents/skills`, `~/.dsh/skills`, `~/.agents/skills`. |
| A builtin could be missing from the table undetected. | A test fails unless every builtin is registered. |

## Test plan

- `pnpm --filter @lody/shared exec vitest run tests/acp-skills.test.ts` — 22 passed.
- `pnpm --filter @lody/components exec vitest run tests/mention-skill-source.test.ts` — 22 passed.
- `pnpm format`, then `pnpm typecheck`, `pnpm lint`, `pnpm lint:i18n`, `pnpm check:code-collab-imports`, `pnpm check:platform-boundaries`, `pnpm check:public-boundary` — all clean.
- `packages/components` has 19 test files / 72 tests failing on this branch, all pre-existing: the same 19/72 fail on an unmodified tree at the same base commit (`use-chat-landing-defaults`, `vscode-diff-theme-registration` and similar jsdom teardown failures, none skill-related).
- Regression check: reverting only `mention-skill-source.tsx` to the base revision makes `returns no whitelist for an unregistered agent type instead of an empty one` fail while the other 21 cases pass, so that test fails on the old behavior through the production owner rather than a fixture.
- Scan-set effect, diffed on the built aggregates: project gains `.grok/skills`, `.kimi-code/skills`, `.cursor/skills`, `.dsh/skills` and loses `.kimi/skills`; global gains `~/.grok/skills`, `~/.kimi-code/skills`, `~/.dsh/skills`. Only `.cursor/skills` is new to the shared scan set — the Claude dirs were already in it via the `claude` entry.
- Replaced one stale assertion that pinned `kimi-code-cli` to no global dirs. It now uses `promptscript`, which keeps the same shape under test (project dirs, zero global dirs) rather than duplicating the adjacent `cline` case.
- Not done: no desktop run against live sessions. Dir facts come from each engine's source or docs.

- **Review focus:** the entries in `packages/shared/src/acp/skills.ts`, and the `null`-vs-empty whitelist change in `getAllowedSkillMentionDirs`, which reaches every unregistered agent, not only grok.
- **Decisions to challenge:** (1) grok's entry includes the Claude and Cursor compat roots. They are on by default in the engine but user-disablable, and the table cannot express "configurable"; including them also adds `.cursor/skills` to the shared project scan set. (2) `deepseek`'s dirs hold for the standard and code presets; `minimal` mounts no skill plugin and `cordis` adds a packaged root, and this per-agent table cannot vary by preset. (3) #150 also proposed making `.agents/skills` an unconditional baseline for every agent. Not done: it contradicts the invariant in `mentions/AGENTS.md` that the dir is "a provider-specific alias, not a universal fallback".
- **Plausible failures / evidence gaps:** Kimi's user dirs follow `KIMI_CODE_HOME`, so the hardcoded `~/.kimi-code` misses a relocated data root.

### Authoring context

- **User goal / directives:** After the maintainer invited a PR on #150, the repo user asked whether a cleaner fix existed than patching three table rows, then asked to open the PR and resolve objections in review.
- **Constraints / non-goals:** No behavior change for agents already correctly mapped; no new dir claim without a source in that engine's own source or docs.
- **Risk-bearing decisions:** Widening the `$` menu for unregistered agents (more candidates offered, never fewer); dropping `.kimi/skills`, which removes it from the shared project scan set so a project that still has one stops being surfaced, matching what the engine reads.
- **Destructive or irreversible behavior:** None. No migration, no writes to user data. The version bump only discards a cache that is rebuilt on demand.
- **Deliberately not done or tested:** The table stores each engine's dirs once per agent-type alias, so ~94 keys hold ~69 distinct dir tuples. Splitting engine facts from the alias mapping would remove that duplication but touches every row, so it is left out of a fix PR and can be raised separately. Kimi's `extra_skill_dirs` and built-in tiers are not modeled.
- **Unknowns / confidence:** High confidence in the pi, grok, kimi, and deepseek mappings — grok's and kimi's come from engine source, deepseek's from its plugin config defaults plus Lody's preset composition, pi's from its documented locations. The `null` contract is already stated at the call site (`null = show all`). Lower confidence on whether the grok compat roots belong in the table given they are configurable.

### DIFF
--- a/packages/components/src/components/mentions/AGENTS.md
+++ b/packages/components/src/components/mentions/AGENTS.md
@@ -23,6 +23,9 @@ Product-level mention sources built on `src/ui/mention`.
   it. The scanner handles flat `~/<agent>/skills/<skill>/SKILL.md` and catalog
   `~/<agent>/skills/<category>/<skill>/SKILL.md`; paths outside the provider's
   registered roots (e.g. plugin caches) appear only once their dirs are added.
+- A registered entry mapping to no dir (`deepagents`) keeps its empty
+  whitelist; an unregistered agent type gets `null` instead, because an empty
+  `Set` here filters every candidate out rather than restricting nothing.
 - Before send, known `$` skill tokens are expanded in prompt text to
   `use /token [Skill Path](path)`. Project skills use their project-relative
   `SKILL.md` path; home-scoped (`global` + `system`) skills use the CLI-provided
--- a/packages/components/src/components/mentions/mention-skill-source.tsx
+++ b/packages/components/src/components/mentions/mention-skill-source.tsx
@@ -5,6 +5,7 @@ import {
   getRegisteredGlobalSkillDirs,
   getRegisteredSkillDirs,
   getRegisteredSystemSkillDirs,
+  isRegisteredSkillAgentType,
   type AgentConfigCliType,
   type ProjectSkill,
   type ProjectSkillScope,
@@ -167,6 +168,11 @@ export function getAllowedSkillMentionDirs(
   skillAgent: { cliType?: AgentConfigCliType; agentType?: string } | undefined
 ): ReadonlySet<string> | null {
   if (!skillAgent?.cliType || !skillAgent.agentType) {
+    return null;
+  }
+  // Unregistered types have no known dirs; an empty whitelist would hide every
+  // skill, so they get none. A registered one keeps its whitelist when empty.
+  if (!isRegisteredSkillAgentType(skillAgent.agentType)) {
     return null;
   }
   const agent = { cliType: skillAgent.cliType, agentType: skillAgent.agentType };
--- a/packages/components/tests/mention-skill-source.test.ts
+++ b/packages/components/tests/mention-skill-source.test.ts
@@ -1,8 +1,14 @@
 import { describe, expect, it } from 'vitest';
-import { applyTextRewrites, type ProjectSkill, type ProjectSkillGroup } from '@lody/shared';
+import {
+  applyTextRewrites,
+  DEFAULT_AGENTS_GLOBAL_SKILL_DIR,
+  type ProjectSkill,
+  type ProjectSkillGroup,
+} from '@lody/shared';
 import {
   buildSkillMentionItems,
   buildSkillMentionRewrites,
+  getAllowedSkillMentionDirs,
   getSkillMentionToken,
   hydrateSkillMentionsFromText,
   mergeMentionSkillState,
@@ -111,6 +117,45 @@ describe('buildSkillMentionItems', () => {
         .map((i) => i.dir)
         .sort()
     ).toEqual(['.agents/skills', '.claude/skills', '~/.claude/skills']);
+  });
+});
+
+describe('getAllowedSkillMentionDirs', () => {
+  it('returns no whitelist when the agent is unknown or only partly identified', () => {
+    expect(getAllowedSkillMentionDirs(undefined)).toBeNull();
+    expect(getAllowedSkillMentionDirs({ cliType: 'custom' })).toBeNull();
+    expect(getAllowedSkillMentionDirs({ agentType: 'claude' })).toBeNull();
+  });
+
+  it('returns no whitelist for an unregistered agent type instead of an empty one', () => {
+    // An empty-but-present whitelist filters every candidate out; `null` keeps
+    // them. Custom agents get generated `custom-<uuid>` types.
+    const allowed = getAllowedSkillMentionDirs({ cliType: 'custom', agentType: 'custom-2f1a' });
+    expect(allowed).toBeNull();
+
+    const items = buildSkillMentionItems(GROUPS);
+    expect(selectSkillMentionCandidates(items, '', allowed).map((i) => i.token)).toEqual([
+      'claude-only',
+      'code-review',
+      'deep-research',
+      'global-only',
+      'qwen-only',
+    ]);
+  });
+
+  it('keeps an explicit empty whitelist for a registered agent that reads no skills', () => {
+    const allowed = getAllowedSkillMentionDirs({ cliType: 'registry', agentType: 'deepagents' });
+    expect(allowed).not.toBeNull();
+    expect([...(allowed ?? [])]).toEqual([]);
+    expect(selectSkillMentionCandidates(buildSkillMentionItems(GROUPS), '', allowed)).toEqual([]);
+  });
+
+  it('offers ~/.agents/skills to grok, pi, and kimi', () => {
+    for (const agentType of ['grok', 'pi-acp', 'kimi-code']) {
+      expect(getAllowedSkillMentionDirs({ cliType: 'registry', agentType })).toContain(
+        DEFAULT_AGENTS_GLOBAL_SKILL_DIR
+      );
+    }
   });
 });
 
--- a/packages/shared/src/acp/skills.ts
+++ b/packages/shared/src/acp/skills.ts
@@ -85,6 +85,13 @@ export const ACP_SKILL_DIRS_BY_AGENT_TYPE: Record<string, SkillDirsByAgentType> 
   crush: skillDirs(['.crush/skills'], ['~/.config/crush/skills']),
   cursor: skillDirs([DEFAULT_PROJECT_SKILL_DIR], ['~/.cursor/skills']),
   deepagents: skillDirs([], []),
+  /* `dsh-skill-filesystem`, mounted by the standard and code presets with no
+     roots configured, discovers the project and user defaults. `minimal` and
+     `cordis` differ, which this per-agent table cannot express. */
+  deepseek: skillDirs(
+    [DEFAULT_PROJECT_SKILL_DIR, '.dsh/skills'],
+    ['~/.dsh/skills', DEFAULT_AGENTS_GLOBAL_SKILL_DIR]
+  ),
   devin: skillDirs(
     ['.devin/skills', '.windsurf/skills'],
     [DEFAULT_AGENTS_GLOBAL_SKILL_DIR, '~/.config/devin/skills']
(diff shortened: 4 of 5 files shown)


wibus-wee (approved):

PR #245, opened 2026-08-31
fix(components): localize PR repair prompts

### LINKED ISSUE #244: [Bug] PR repair actions send English prompts in Chinese locale
### Affected area

GitHub / code review

### Installation method

Built from source

### Lody version or commit

d6a23ed27c9882388f9219b3c5587f65ac4bdc11

### Operating system

macOS 27.0 arm64

### Agent or runtime

The issue is independent of the selected agent runtime because the desktop client constructs the prompt before dispatch.

### What happened?

With the Lody interface set to Simplified Chinese, the PR quick-action labels are translated, but the prompt submitted by “Resolve conflicts” is still English. The “Fix CI errors” action uses the same hard-coded-English prompt path, so it has the same locale mismatch.

This makes a Chinese session unexpectedly switch to an English task instruction even though nearby PR actions already follow the UI locale.

Image

### What did you expect?

PR repair quick actions should construct their complete task prompts using the active UI locale. English behavior and the existing CI metadata, filtering, and length safeguards should remain unchanged.

### How can we reproduce it?

1. Build and start Lody from the referenced commit.
2. Set the interface language to Simplified Chinese.
3. Open a session associated with a GitHub pull request that has merge conflicts.
4. Click “解决冲突”.
5. Observe that the submitted task prompt begins with “Resolve the merge conflicts”.
6. On a pull request with failed checks, click “修复 CI” and observe the same English-prompt mismatch.

### How often does it happen?

Every time

### Additional context

The button labels use i18n, while the prompt builders in `session-pr-prompts.ts` contain hard-coded English strings. This separates the visible locale from the text actually dispatched to the agent.

### Before submitting

ImSingee:
## Related issue

Closes #244

## Problem / pressure

The PR repair quick-action labels follow the active UI locale, but the “Resolve conflicts” and “Fix CI errors” builders hard-code English. In a Simplified Chinese session, clicking either localized action therefore dispatches an English task prompt and breaks the language continuity seen in nearby actions such as Create PR.

## Summary

- Pass the active i18n translator into both PR repair prompt builders.
- Add complete English and Simplified Chinese resources for conflict resolution and CI repair prompts.
- Keep locale resources as the single source of prompt text and rely on the configured English language fallback instead of duplicated TypeScript templates.
- Preserve failed-check classification, the 12-check cap, the 6,000-character limit, and the warning that check metadata is untrusted.
- Cover English compatibility, Simplified Chinese output, omitted-check localization, and post-interpolation truncation.

## Before / after

| Before | After |
| ------ | ----- |
| Chinese PR repair buttons dispatch English prompts. | PR repair prompts follow the active English or Simplified Chinese UI locale. |
| CI metadata safeguards exist only inside a hard-coded English prompt. | The same safeguards are retained across localized prompt templates. |

## Test plan

- `corepack pnpm test:ci` — passed the full workspace matrix on the initial implementation, including 2,934 component tests, 2,444 CLI tests with 4 skipped, and 54 Electron tests.
- Review follow-up — reran the full Components suite with 2,934 tests passed and the targeted `session-pr-prompts.test.ts` suite with 7 tests passed.
- Component and CLI TypeScript checks — passed; the review follow-up also passed the complete Components TypeScript check.
- `pnpm lint:i18n` — passed with 2,851 keys.
- Type-aware Oxlint — 0 errors; Prettier check and `git diff --check` passed.
- Code Collab import and public-boundary guards — passed.

- **Review focus:** Inspect `session-pr-prompts.ts`, both locale resources, and the dispatch call sites to verify every agent-visible PR repair instruction follows the active locale without duplicating templates in TypeScript.
- **Decisions to challenge:** Confirm that injecting `TFunction` into these pure builders while keeping locale JSON as the only prompt-text source is the appropriate boundary for locale-aware prompt construction.
- **Plausible failures / evidence gaps:** Check interpolation of untrusted CI metadata, omitted-check text, and truncation after localization; unit and full workspace tests cover these paths, but no packaged desktop click-through was run.

### Authoring context

- **User goal / directives:** Make the Chinese “Resolve conflicts” action send a Chinese prompt like Create PR, and apply the same correction to Fix CI; prior maintainer agreement on this scope and approach was confirmed by the author.
- **Constraints / non-goals:** Keep English behavior and all GitHub/CI selection semantics unchanged; do not alter agent selection, PR operations, or failure classification.
- **Risk-bearing decisions:** Dynamic check names and URLs remain verbatim untrusted data, while only surrounding instructions are localized; the 6,000-character cap is applied after interpolation.
- **Destructive or irreversible behavior:** The change performs no migration, deletion, or external write at runtime; rollback is a normal revert of these isolated commits.
- **Deliberately not done or tested:** No packaged Electron UI click-through or live GitHub repair action was run because builder tests and the full component suite exercise the dispatched text path without mutating a real PR.
- **Unknowns / confidence:** Confidence is high for the supported English and Simplified Chinese locales; locale JSON is the single prompt source, `fallbackLng: 'en'` covers runtime gaps, and `lint:i18n` prevents checked-in key drift.


packages/components/src/components/sessions/session-pr-prompts.ts
```diff
@@ -7,6 +8,32 @@ import type {
 const MAX_FAILED_RUNS = 12;
 const MAX_PROMPT_LENGTH = 6_000;
 
+const FIX_CI_ERRORS_PROMPT = [
```
wibus-wee: Thanks for your contribution! 

But I don’t think the duplicated English templates in TS are necessary here. 

These TS constants look like a historical fallback pattern. Since the locale resources are statically bundled, `fallbackLng: 'en'` already covers missing translations, and `check-i18n` validates both locales, the duplicated TS templates don’t seem necessary. 

Could we clean them up as part of this change to avoid future drift?

wibus-wee (changes requested):

packages/components/src/components/sessions/session-pr-prompts.ts
ImSingee: Addressed in 9de41e9. I removed the duplicated TypeScript English templates; the builders now call t(key, interpolation) only, leaving locale JSON as the single source of prompt text. Runtime fallbackLng: en and lint:i18n cover fallback behavior and key completeness. The targeted suite passed 7/7, the full Components suite passed 2,934/2,934, and the Components typecheck and i18n validation passed.

wibus-wee (approved):
LGTM

PR #285, opened 2026-09-02
fix(components): restore mobile settings source

### LINKED ISSUE #263: [Bug] Mobile settings back resets Projects tab to Chat
### Affected area

Other open-source component (`packages/components` shared mobile workspace UI and routing)

### Installation method

Other (native mobile shell consuming the shared open-source components)

### Lody version or commit

8a674e5

### Operating system

iOS native shell on iPhone; the exact iOS version and device model were not captured.

### What happened?

Opening Settings from the mobile workspace home while the Projects tab is selected, then pressing the Settings back button, returns to the Chat tab. This loses the user's active workspace context and makes a routine settings round trip feel like an unintended tab switch.

The behavior is caused by the shared open-source route integration: the mobile Settings entry does not preserve the current `/chat?context=local` or `/chat?context=github` location, and the top-level Settings back action always navigates to a context-free `/chat` route. `ChatLanding` then mounts with its default Chat tab.

### What did you expect?

Leaving Settings should restore the workspace home tab and project sub-tab from which Settings was opened. A user who entered Settings from Projects / Local should return to Projects / Local, and the same should hold for Projects / GitHub. Existing nested Settings navigation should continue to return to the Settings category list before exiting Settings.

### How can we reproduce it?

1. Open the mobile workspace home.
2. Switch from Chat to the Projects tab.
3. Select either the Local or GitHub project sub-tab.
4. Tap the Settings button in the upper-right corner.
5. Press the Settings back button.
6. Observe that the workspace home opens on Chat instead of the previously selected Projects tab.

### How often does it happen?

Every time

### Additional context

Suggested scope: preserve the complete in-app source path when the mobile Settings route is opened, and have the top-level Settings back action prefer that validated source path over the context-free Chat fallback. Add a regression test for source-aware Settings exit while retaining the current nested-settings-to-list behavior.

This report concerns the public shared component and routing implementation, not private mobile-shell source or a hosted service. The screenshot used during local diagnosis contained repository identity details and is intentionally not attached.

### Before submitting

MichaelYuhe:
## Related issue

Closes #263

## Problem / pressure

Mobile workspace Settings discarded the source query when opened from Projects, and the top-level Back action always navigated to context-free Chat. A Settings round trip therefore changed Projects / Local or Projects / GitHub into the Chat tab.

## Summary

- Carry the complete current in-app path in the Settings `from` search parameter when mobile workspace home opens Settings.
- Make top-level Settings Back restore a validated source path while preserving the existing nested-settings-to-list behavior and direct-entry Chat fallback.
- Add focused Local, GitHub, nested-page, missing-source, and unsafe-source regression coverage.
- Record the mobile Settings navigation invariant in the scoped contributor guidance.

## Before / after

| Before | After |
| ------ | ----- |
| Settings Back from Projects / Local or Projects / GitHub opened context-free Chat. | Settings Back restores the exact originating Projects context and query. |
| Nested Settings Back returned to the Settings list. | Nested Settings Back still returns to the Settings list before leaving Settings. |

## Test plan

- `npx --yes vitest@3.2.4 run --config /tmp/lody-vitest.config.mjs` — 4 tests passed.
- `npx --yes oxlint@1.61.0 <changed TypeScript files>` — 0 errors; 7 pre-existing warnings in `chat-landing.tsx`.
- `git diff --check upstream/main...HEAD` — passed.
- Full workspace typecheck and test suite were not run because this checkout has no `corepack`, `pnpm`, or installed workspace dependencies; PR CI should run the canonical checks.

- **Review focus:** Check `chat-landing.tsx`, `settings.tsx`, and `settings-navigation.ts` for preservation and restoration of the full mobile Projects source URL.
- **Decisions to challenge:** Verify that nested Settings Back must outrank the saved source while top-level Back may restore any validated in-app absolute path.
- **Plausible failures / evidence gaps:** The focused pure tests pass, but the full router integration and workspace suite were left to CI because dependencies are unavailable in this checkout.

### Authoring context

- **User goal / directives:** Fix Issue #263 after a maintainer requested the PR so mobile Settings returns users to their selected Projects context.
- **Constraints / non-goals:** Keep the change in public shared components, preserve nested Settings behavior, and avoid private native-shell changes.
- **Risk-bearing decisions:** Store the full current path in `from` and accept only absolute non-protocol-relative in-app paths before restoring it.
- **Destructive or irreversible behavior:** None; navigation state only, with context-free Chat retained as the recovery fallback.
- **Deliberately not done or tested:** No native-shell changes or broad workspace tests; the isolated regression suite and lint completed locally, with canonical checks delegated to CI.
- **Unknowns / confidence:** High confidence in route decision logic; remaining uncertainty is limited to end-to-end rendering in the unavailable native mobile shell.

### DIFF
--- a/packages/components/src/components/chat/chat-landing.tsx
+++ b/packages/components/src/components/chat/chat-landing.tsx
@@ -80,6 +80,7 @@ import { docMetaCacheReadyAtom, sessionMetaCountAtom } from '@/atoms/doc-meta';
 import { localProbeAttemptedAtom, localProbeResultAtom } from '@/atoms/local-probe';
 import { lodyPresenceNowMsAtom, lodyPresenceStatesAtom } from '@/atoms/presence';
 import { buildAgentPrompt } from '@/lib';
+import { getAppCurrentPathWithSearch } from '@/lib/app-location';
 import { isImeComposingKeyboardEvent } from '@/lib/ime';
 import { useNavigate } from '@tanstack/react-router';
 import { activeWorkspaceRuntimeAtom, authTokenAtom, runtimeAtom } from '@/atoms/runtime';
@@ -6415,6 +6416,7 @@ function WorkspaceChatLanding({
             void navigate({
               to: '/$workspaceName/settings',
               params: { workspaceName: workspaceSlug },
+              search: { from: getAppCurrentPathWithSearch() },
             });
           }}
           /* New-chat chip opens the bottom-sheet composer. Stays on the
--- a/packages/components/src/components/mobile/AGENTS.md
+++ b/packages/components/src/components/mobile/AGENTS.md
@@ -105,6 +105,10 @@ embedded` lazy-imported from `../tasks/tasks-workspace.tsx` (`embedded`
   `workspaceJoinRequestsSlot` already owns its card frame, so wrap it in a
   `MobileSettingsSection noCard` plus the standard `mx-3` card inset; mounting
   the slot raw drops both the section rhythm and the shared horizontal edge.
+  Opening the mobile Settings route from workspace home carries the complete
+  in-app path in `from`. Nested settings Back returns to the settings list;
+  top-level Back restores that validated source path (including the Projects
+  Local/GitHub query) and uses context-free Chat only as the direct-entry fallback.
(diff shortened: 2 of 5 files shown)


Leeeon233:
Nice

PR #299, opened 2026-09-02
fix(electron): allow fake-IP proxy answers in the public browser

### LINKED ISSUE #298: [Bug] Public browser blocks every site behind a fake-IP proxy (198.18.0.0/15)
### Affected area

Desktop app

### Installation method

Desktop release

### Lody version or commit

0.89.4 (source commit 896fd1e)

### Operating system

macOS 26.4 arm64

### Agent or runtime

Not agent-specific. Reproduced with Clash-style fake-IP DNS enabled on the machine.

### What happened?

With a fake-IP proxy running (Clash / Clash Verge, Surge, sing-box, Shadowrocket in `fake-ip` DNS mode), every public site opened in the Session Browser fails immediately with:

```
Public browser blocked a non-public destination: 198.18.3.75
```

The public browser resolves the hostname before navigating and rejects any answer that is not a public address. Fake-IP proxies answer every DNS query from the RFC 2544 benchmarking range `198.18.0.0/15` and then intercept the connection by domain, so the resolver answer is always a synthetic `198.18.x.x` address. That range is on the prohibited list in `packages/shared/src/browser-url.ts`, so the navigation is blocked before it starts. Fake-IP mode is the default for several of these tools and is very common, so the public browser is unusable for those users while the proxy is on.

### What did you expect?

Public sites should open normally. A public hostname resolving into `198.18.0.0/15` is the proxy's synthetic answer rather than a reachable private host, so it should not trip the non-public-destination guard. Typing a literal `198.18.x.x` address should remain blocked.

### How can we reproduce it?

1. Enable a fake-IP proxy on the machine (for example Clash with `dns.enhanced-mode: fake-ip`, TUN or system DNS pointed at it) so that `dig example.com` returns an address in `198.18.0.0/16`.
2. Start the Lody desktop app and open a session.
3. Open the Session Browser and enter `https://example.com`.
4. Observe the error `Public browser blocked a non-public destination: 198.18.x.x` and no page load.

### How often does it happen?

Every time

### Relevant log output

```shell
Public browser blocked a non-public destination: 198.18.3.75
```

### Additional context

The guard itself is correct for every other reserved range; only `198.18.0.0/15` collides with fake-IP resolvers. The fix should distinguish a resolver answer in that range from a literal address so the DNS-rebinding protection for loopback, private LAN, link-local, and metadata addresses stays intact.

### Before submitting

EYHN:
## Related issue

Closes #298

## Problem / pressure

The public browser resolves every hostname before navigating and rejects any answer that is not a public address, which defeats DNS rebinding against loopback, private LAN, link-local, and cloud-metadata targets. Fake-IP proxies (Clash / Clash Verge, Surge, sing-box, Shadowrocket in `fake-ip` DNS mode) answer every DNS query from the RFC 2544 benchmarking range `198.18.0.0/15` and intercept the connection by domain. That range is on the prohibited list, so with such a proxy active every public site fails with `Public browser blocked a non-public destination: 198.18.x.x`. Fake-IP mode is the default for several of these tools, so the public browser is unusable for a large group of users while their proxy is on.

## Summary

- Add `classifyResolvedBrowserAddress` in `packages/shared/src/browser-url.ts`. It classifies an address the resolver returned for a public hostname and differs from `classifyBrowserHostname` only for `198.18.0.0/15`, which it treats as public. IPv4-mapped IPv6 answers are unwrapped the same way.
- `PublicBrowserService.resolvePublicHostname` now uses the resolved-address classifier, so the navigation, `webRequest`, and WebSocket policy paths all accept fake-IP answers while still blocking every other non-public range.
- Literal `198.18.x.x` addresses typed into the address bar or used as a page subresource remain prohibited; `classifyBrowserHostname` and `parseBrowserAddress` are unchanged for them.
- Shared the IPv4-mapped IPv6 unwrapping between `classifyIpv6` and the new classifier instead of duplicating it.
- Tests cover the `/15` boundary, the mapped-IPv6 form, and that loopback, private LAN, link-local, metadata, and documentation ranges are still blocked when resolved.
- Recorded the exception next to the dual-engine invariant in `packages/components/src/components/sessions/AGENTS.md`.

## Before / after

| Before | After |
| ------ | ----- |
| With a fake-IP proxy on, `https://example.com` in the Session Browser fails with `Public browser blocked a non-public destination: 198.18.3.75`. | The page loads; the proxy intercepts the `198.18.x.x` connection and forwards it by domain as it does for every other application. |
| A hostname resolving to `127.0.0.1`, `10.x`, `169.254.169.254`, or `198.51.100.x` is blocked. | Unchanged: still blocked. |
| Typing `http://198.18.3.75/` is rejected as a reserved address. | Unchanged: still rejected. |

## Test plan

- `pnpm --filter @lody/shared exec vitest run tests/browser-url.test.ts`: 36 passed, including the new `classifyResolvedBrowserAddress` cases and the literal-address prohibition case.
- `pnpm --filter @lody/shared typecheck` and `pnpm --filter @lody/electron typecheck` (node and web): clean.
- `oxlint --type-aware` on the changed files: no errors; the remaining warnings are pre-existing lines not touched here. `eslint` in `apps/electron` on the changed service: clean.
- `pnpm --filter @lody/electron test`: 54 passed.
- `pnpm check:public-boundary`: passed. `prettier --check` on the changed files: clean.
- Skipped: the full `pnpm check` (root typecheck and the complete test matrix) and a manual desktop run behind a live fake-IP proxy. The behavior change is fully covered by the shared unit tests; the Electron service change is a one-symbol swap on a code path that cannot run under `node --test` because it imports `electron`.

- **Review focus:** `packages/shared/src/browser-url.ts` (`classifyResolvedBrowserAddress`, `isBenchmarkingIpv4`, `mappedIpv4`) and the single call-site change in `apps/electron/src/main/services/public-browser-service.ts`; confirm every path that checks resolver answers goes through `resolvePublicHostname`.
- **Decisions to challenge:** exempting the whole `198.18.0.0/15` instead of the more common `198.18.0.0/16` (sing-box defaults to the `/15`); keeping the exemption resolver-only rather than also allowing literal addresses; not adding a user-facing hint about fake-IP proxies in the error message.
- **Plausible failures / evidence gaps:** a network that illegitimately uses `198.18.0.0/15` for internal hosts would now be reachable via DNS rebinding when no fake-IP proxy is active; not verified against a live fake-IP proxy in the desktop app, only via unit tests of the classifier.

### Authoring context

- **User goal / directives:** make the public browser work behind fake-IP proxies by treating `198.18.0.0/15` differently from other reserved ranges, and open a PR that explains why.
- **Constraints / non-goals:** keep the DNS-rebinding guard for every other range; keep literal reserved addresses prohibited; no changes to Managed Preview, the dual-engine routing, or the error UI.
- **Risk-bearing decisions:** narrowing the security policy for one non-globally-routed range; the exemption is resolver-only and documented as a deliberate exception in the sessions `AGENTS.md`.
- **Destructive or irreversible behavior:** none; no data, migration, or configuration changes.
- **Deliberately not done or tested:** no live fake-IP desktop run and no full-root `pnpm check`; the classifier is exhaustively unit tested and the service change is a one-symbol swap on a typechecked path.
- **Unknowns / confidence:** high confidence in the classifier; moderate residual risk only for networks that misuse the benchmarking range internally, which is uncommon and already outside the range's specification.

### DIFF
--- a/apps/electron/src/main/services/public-browser-service.ts
+++ b/apps/electron/src/main/services/public-browser-service.ts
@@ -6,7 +6,7 @@ import {
   type ElectronPublicBrowserResult,
   type ElectronPublicBrowserState
 } from '@lody/shared/electron-ipc'
-import { classifyBrowserHostname, parseBrowserAddress } from '@lody/shared/browser-url'
+import { classifyResolvedBrowserAddress, parseBrowserAddress } from '@lody/shared/browser-url'
 import { formatUnknownError } from '../utils'
 import { isNavigationAbortError, mergePublicBrowserState } from './public-browser-state'
 
@@ -54,7 +54,7 @@ const resolvePublicHostname = async (browserSession: Session, hostname: string):
     throw new Error(`Unable to resolve public browser host: ${hostname}`)
   }
   const blocked = endpoints.find(
-    (endpoint) => classifyBrowserHostname(endpoint.address) !== 'public'
+    (endpoint) => classifyResolvedBrowserAddress(endpoint.address) !== 'public'
   )
   if (blocked) {
     throw new Error(`Public browser blocked a non-public destination: ${blocked.address}`)
--- a/packages/components/src/components/sessions/AGENTS.md
+++ b/packages/components/src/components/sessions/AGENTS.md
@@ -242,6 +242,10 @@ Session conversation page chain:
   capability (Electron `WebContentsView` today); loopback/private targets use Managed Preview and
   are the only pages eligible for Visual Annotation. Never fall back from a missing public engine
   to iframe, system browser, CLI, Preview Gateway, or a different Machine RPC plane.
+  The public engine resolves every hostname and rejects non-public answers (DNS rebinding), with
+  one deliberate exception: a public name resolving into RFC 2544 `198.18.0.0/15` is a fake-IP
+  proxy's synthetic answer (Clash, Surge, sing-box, Shadowrocket) and is allowed via
+  `classifyResolvedBrowserAddress`; a literal `198.18.x.x` address stays prohibited.
   The composer info-bar Browser action is an explicit candidate-navigation request, not merely a
   panel-open action. It opens the reported candidate even when another page is already visible.
   That click IS the approval for that exact target: a remote route creates (or replaces) its tunnel
--- a/packages/shared/src/browser-url.ts
+++ b/packages/shared/src/browser-url.ts
@@ -48,14 +48,16 @@ const stripIpv6Brackets = (host: string): string => {
     : normalized;
 };
 
-const parseIpv4 = (host: string): [number, number, number, number] | null => {
+type Ipv4Octets = [number, number, number, number];
+
+const parseIpv4 = (host: string): Ipv4Octets | null => {
   const parts = host.split('.');
   if (parts.length !== 4) return null;
   const octets = parts.map((part) => Number(part));
   if (octets.some((octet) => !Number.isInteger(octet) || octet < 0 || octet > 255)) {
     return null;
   }
-  return octets as [number, number, number, number];
+  return octets as Ipv4Octets;
 };
 
 const parseIpv6 = (host: string): number[] | null => {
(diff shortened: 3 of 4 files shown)


Leeeon233:
Oh! Thanks!

PR #309, opened 2026-09-02
fix(components): preserve mention caret during IME composition

### LINKED ISSUE #307: [Bug] Mention chip obscures the caret and disrupts CJK IME editing
### Affected area

Desktop app

### Installation method

Built from source

### Lody version or commit

94bede9

### Operating system

macOS arm64

### Agent or runtime

Any composer session using @ session mentions

### What happened?

After inserting a session mention into the Chat composer, moving the caret to the text immediately before or after the mention can produce editing glitches. When the caret lands inside an opaque mention chip, the native caret is hidden. If the mention is at the beginning of the draft, moving left can make the caret appear to disappear entirely. During Chinese/Japanese/Korean IME composition near a mention, the transient preedit text can become visually misaligned or duplicate/shift around the chip.

### What did you expect?

Mention chips should remain visually styled without covering the native caret. The caret should stay visible at both mention boundaries, and IME preedit text should remain aligned while the committed mention range stays intact.

### How can we reproduce it?

1. Open a Chat composer with session mentions enabled.
2. Type ordinary text, insert a session with `@`, and leave text on either side of the chip.
3. Move the caret with the arrow keys or pointer to just before and just after the mention; also try a mention at offset 0 and move left.
4. Repeat with a Chinese IME active and type a short preedit string adjacent to the mention.
5. Observe the hidden/displaced caret or the misaligned composition text.

https://github.com/user-attachments/assets/77ca4c4a-08d1-4735-8f9f-16bd5fdff558

### How often does it happen?

Every time

### Relevant log output

```text

```

### Additional context

The issue is isolated to the live composer overlay made of a native textarea plus mention mirrors. It does not change the text sent to the agent; it affects caret visibility, atomic mention editing, and IME rendering.

### Before submitting

jeffwcx:
## Related issue

Closes #307

## Problem / pressure

The Chat composer uses a native textarea plus an opaque mirror for mention chips. A collapsed caret moved into or beside a chip can be covered or disappear, and a CJK IME's transient preedit text can become offset from the committed mention range. The failure is most visible when a mention starts the draft or when a long mention wraps.

## Summary

- Treat collapsed carets inside committed mentions as atomic and snap them to the nearest boundary, while leaving non-collapsed selections native.
- During IME composition, render the transient textarea value and move mention ranges through the transient edit so both mirrors stay aligned.
- At a mention boundary, raise the textarea above the chip and make its text fill transparent while the background mirror carries the glyphs; this keeps the native caret visible without painting a second caret.
- Keep readonly boundary handling visual-only, ignore mention clicks during composition, and add regression coverage for navigation, wrapping, readonly inputs, and CJK preedit alignment.

## Before / after

| Before | After |
| ------ | ----- |
| Chip mirror could cover the native caret at a mention edge or at offset 0. | The native caret remains visible at both boundaries without a synthetic caret overlay. |
| CJK IME preedit text could shift or duplicate around an opaque chip. | Transient preedit text and mention ranges share the same mirror coordinate space. |
| Moving through a chip could split or silently decommit the mention range. | Collapsed carets stay outside committed ranges; ordinary selections still work. |

## Test plan

- `pnpm --filter @lody/components exec vitest run tests/mention-navigation.test.tsx` — 16/16 passed.
- `pnpm --filter @lody/components exec vitest run` — 420 test files and 3022 tests passed.
- `pnpm --filter @lody/components typecheck` — passed.
- Prettier check on all four changed files — passed.
- `git diff --check` — passed.
- Chromium manual layout check covered a wrapped mention at 220px; boundary carets measured as one line high on their respective fragments.
- Full workspace `pnpm check` was not rerun; this PR is limited to the components package and its focused/full package suites.

- **Review focus:** Inspect `mention-input.tsx` and `mention-highlighter.tsx` together; the fix depends on their shared caret-layer and transient-range coordinate contract.
- **Decisions to challenge:** Verify snapping only collapsed carets, raising the textarea only at boundaries, and using the background mirror for text are correct across wrapped chips and readonly inputs.
- **Plausible failures / evidence gaps:** Browser coverage is Chromium-based; WebKit/Firefox IME behavior and unusual grapheme edits remain the main residual compatibility risks.

### Authoring context

- **User goal / directives:** Keep the Chat composer caret visible around `@` session chips and prevent CJK IME text from becoming visually misaligned.
- **Constraints / non-goals:** Preserve the native textarea, mention payload/range semantics, chip styling, autocomplete behavior, and ordinary text selections; no protocol or persistence changes.
- **Risk-bearing decisions:** Mention ranges are treated as atomic only for collapsed caret placement; transient IME edits are mapped to the same ranges instead of hiding chips or committing partial text.
- **Destructive or irreversible behavior:** None. The change is presentation and selection handling only; it performs no migration, deletion, overwrite, or external data write.
- **Deliberately not done or tested:** Full workspace `pnpm check` and native WebKit/Firefox IME runs were not performed; package-wide tests, typecheck, formatting, and Chromium layout checks cover the touched surface.
- **Unknowns / confidence:** High confidence for the tested Chromium/Electron path and jsdom regressions; moderate confidence for other browser IME implementations until they are exercised.

### DIFF
--- a/packages/components/src/ui/mention/AGENTS.md
+++ b/packages/components/src/ui/mention/AGENTS.md
@@ -24,6 +24,18 @@ Shared mention primitive used by composer autocomplete surfaces.
   commits a highlighted non-navigation item the same way Enter does.
   Shift+Tab still closes the menu so the composer mode-cycle binding
   is not stolen.
+- A committed mention is an atomic editing range. A collapsed caret placed
+  inside it by pointer/focus/selection changes snaps to the nearest boundary;
+  otherwise the chip mirror hides the native caret and the next edit silently
+  decommits the range. Non-collapsed selections remain native so copy and
+  whole-region edits can span mentions. At a mention boundary, the textarea is
+  raised above the opaque chip and its text fill is made transparent while the
+  background mirror carries the glyphs; this leaves the native caret visible
+  without painting a second caret. Readonly inputs still apply this visual
+  boundary constraint. IME composition is the offset exception: its transient
+  caret must not snap against the still-committed ranges, and both mirrors map
+  those ranges through the transient edit so the chip neither disappears nor
+  exposes a duplicated suffix.
 - The pop-back itself is `context.onNavigateBack()`, owned by the root next to
   `onMentionAdd`: it has to interleave the controlled value commit with caret
   restoration, so a menu's own Back affordance calls it rather than restaging the
--- a/packages/components/src/ui/mention/mention-highlighter.tsx
+++ b/packages/components/src/ui/mention/mention-highlighter.tsx
@@ -15,8 +15,10 @@ type HighlighterElement = HTMLDivElement;
  * glyphs — the plain highlight. `chip` sits *over* it and paints opaque chips
  * that hide the raw characters they replace, which is the only way to show an
  * icon and a chip-specific text colour without giving up the native textarea.
- * Both mirrors render the identical character stream, so they agree on every
- * line break and the caret stays aligned with both.
+ * At a mention boundary MentionInput temporarily raises the textarea above
+ * `chip`, so `background` carries the text while the native caret remains on
+ * top. Both mirrors render the identical character stream, so they agree on
+ * every line break and the caret stays aligned with both.
  */
 type MentionHighlighterLayer = 'background' | 'chip';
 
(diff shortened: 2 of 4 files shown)


Leeeon233:
Thanks!

PR #312, opened 2026-09-02
fix(components): mention a dropped folder as a path instead of a failed attachment

### LINKED ISSUE #311: [Bug] Dropping a folder onto the chat input shows a failed attachment instead of a path
### Affected area

Desktop app

### Installation method

Built from source

### Lody version or commit

896fd1e

### Operating system

macOS 26 arm64

### Agent or runtime

Not runtime-specific; the failure is in the shared composer before anything is sent.

### What happened?

Dragging a folder from Finder onto the chat input (session view or the new-chat landing) adds an attachment chip that immediately shows as failed. The folder cannot be uploaded, so the drop produces nothing usable and the user has to remove the failed chip.

### What did you expect?

The folder should become a path reference in the prompt, like `@/Users/me/project/src`, so the agent can be pointed at a directory by dropping it.

### How can we reproduce it?

1. Start the desktop app from source (`pnpm start:local`) and open any local-project conversation or the new-chat landing.
2. Drag a folder from Finder onto the composer or the conversation area.
3. Observe a file attachment chip appear and switch to the failed state.

### How often does it happen?

Every time

### Relevant log output

```shell
No log output; the failure is visible in the composer's attachment strip.
```

### Additional context

The drop handler in `packages/components/src/lib/file-drop.ts` treats every `DataTransferItem` of kind `file` as an upload candidate. A dropped directory arrives as a `File` with an empty type, so it is queued as an attachment and its read fails. Electron exposes `webUtils.getPathForFile`, which can name the folder's path instead.

### Before submitting

EYHN:
## Related issue

Closes #311

## Problem / pressure

Dropping a folder from the OS onto the composer produced a failed attachment chip. Chromium delivers a dropped directory as a `File` with an empty type and the folder's name, and `lib/file-drop.ts` collected every `DataTransferItem` of kind `file` as an upload candidate, so the folder was queued as an attachment and its read failed. The user's intent when dropping a folder is to point the agent at that directory, which needs a path in the prompt, not an upload.

## Summary

- `lib/file-drop.ts`: `readDroppedTransfer` splits one drop into `files` and `directories` using `webkitGetAsEntry().isDirectory`; directories never reach the attachment queue.
- `lib/dropped-local-path.ts` (new): reads the folder's absolute path through the Electron preload's `webUtils.getPathForFile` and normalizes it (forward slashes, no trailing slash). Returns null on web/mobile or an older preload, where a directory drop then inserts nothing.
- `combined-mention-textarea.tsx`: the `mentionActionsRef` handle gains `insertPathMention`, writing `@<path>` plus a committed range with kind `dir`, the same artefact a menu commit produces, so the chip renders and the token is not sent as a bare word.
- `chat-composer.tsx`: new `onDirectoryDrop` prop; the session input area and the landing view turn dropped directories into path mentions. The session page's whole-conversation drop zone routes directories the same way.
- Mentions and chat `AGENTS.md` record the invariant.

The mention keeps the absolute path deliberately: the agent resolves it from any cwd (worktree, another project, the home folder), and the user reads exactly what was dropped.

## Before / after

| Before | After |
| ------ | ----- |
| Folder drop adds an attachment chip that shows as failed; nothing usable lands in the prompt. | Folder drop inserts `@/absolute/path/to/folder ` as a directory mention with a folder chip; regular files still upload as attachments. |

## Test plan

- `corepack pnpm --filter @lody/components typecheck` passes.
- `vitest run` on `tests/file-drop.test.ts`, `tests/dropped-local-path.test.ts`, `tests/path-mention-insert.test.tsx`, plus the existing `tests/mention-*` and `tests/session-mention*` suites: 20 files, 175 tests pass.
- `oxlint` on the changed files reports no new findings (existing `no-use-before-define` warnings in `chat-landing.tsx` and `session-chat-interface.tsx` are untouched).
- Skipped: a manual drag-and-drop pass in the running desktop app; `check:public-boundary` was not run to completion because the `acp-extension-codex`/`grok` submodules are not checked out in this environment.

- **Review focus:** `packages/components/src/lib/file-drop.ts` (the file/directory split and the `files` fallback) and `combined-mention-textarea.tsx` `insertPathMention` (range value matches `toFileCandidate`: `dir` payload keeps the trailing slash, text does not).
- **Decisions to challenge:** keeping the mention absolute rather than relative to the session cwd; inserting nothing on web/mobile instead of surfacing a message; leaving plain files on the upload path.
- **Plausible failures / evidence gaps:** `webkitGetAsEntry` returning null for some drag sources would classify a folder as a file again; the running app was not exercised manually in this change.

### Authoring context

- **User goal / directives:** dropping a folder onto the chat input should insert its path, and the path must stay absolute.
- **Constraints / non-goals:** file drops keep uploading as attachments; no new user-visible copy; no cloud or platform-capability changes.
- **Risk-bearing decisions:** the mention range uses kind `dir` with a trailing-slash payload so hydration, chips, and the before-send rewrite treat it like a menu-committed directory.
- **Destructive or irreversible behavior:** none; the change only affects what a drop writes into an unsent draft.
- **Deliberately not done or tested:** no manual desktop run; no relative-path conversion; web/mobile directory drops are silently ignored because no path bridge exists there.
- **Unknowns / confidence:** high confidence in the unit-tested pieces; medium on drag sources outside Finder/Explorer that may not expose the entry API.


Leeeon233:
Please resolve the conflicts~

PR #324, opened 2026-09-03
feat(cli): warn when the agent's run config differs from the request

zxch3n:
Stacked on #316 — review that one first; the base flips to `main` once it merges.

## Problem

`applyAcpSessionRunConfig` decided a user-visible `agent_warning` from a REJECTION, then suppressed it for exactly the four controls users most need told about — model, reasoning effort, Fast, Plan — on Codex and Claude (`suppressKnownRunConfigWarnings && isKnownRunConfigOption(...)`). The suppression existed because those calls fail routinely when the turn config is re-sent every turn.

A rejection is the wrong signal, in both directions:

- **False negative.** Codex `applyFastModeChange` sets `fastModeEnabled` unconditionally — it ACCEPTS `fast-mode` on a model with no fast speed tier. Nothing throws; `createSessionConfigOptions` then omits the option and `resolveFastServiceTier` returns `null`, so the turn runs at normal speed and the user is told nothing. Even without the suppression, no rejection ever existed to report.
- **False positive.** A rejected value that was already effective changed nothing, and was reported.

## Fix

Key the notice off **divergence**: after applying the turn's config, compare each requested selection against the state the agent itself publishes (`runtimeConfigPatch`, already computed from `agentClient.getConfigOptions()`), and warn only where they disagree.

- An `on`/`off` select and a boolean toggle express the same choice, so neither shape alone counts as a divergence.
- An explicit `config.modeId`/`modelId` outranks a duplicate config-option entry; losing that precedence contest is not the agent disagreeing, so only the winner is judged.
- Where the agent published no config options at all — or for a sensitive id, which the runtime state deliberately omits — the failed call stays the only signal and is still used.
- `rejectedSelections` (debug diagnostics) is unchanged.

The notice text now says the agent *did not apply* part of the config rather than *rejected* it, which is what the new signal actually means.

Noise is bounded: the composer re-sends the state the agent reported, so an ordinary turn diverges nowhere, and `recordAgentWarning` already deduplicates by message per session — a stuck divergence costs one notice, not one per turn.

## Invariant change

`apps/cli/AGENTS.md` documented the old behavior ("Codex/Claude mismatches for model, reasoning effort, Fast, or Plan are not promoted to visible `agent_warning` notices"). That paragraph is rewritten to state the divergence rule, why a rejection is not it, and to say plainly that a per-agent suppression list must not come back. The Claude Fable `fast=false` no-op exception is preserved — it is skipped before dispatch, so it is judged for neither.

## Tests

`apps/cli/src/session/acp-session-config-applier.test.ts` — the two tests that encoded the suppression are replaced by ones for the new rule: a selection accepted but dropped from the agent's state (the Codex Fast shape, with empty `rejectedSelections`), a rejected selection that was already effective (silent), and the select/boolean equivalence. The four tests covering redaction, non-publishing agents, and other agent types are unchanged and still pass.

`pnpm --filter lody test` (2467), `typecheck`, `lint` all pass.

### DIFF
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -274,12 +274,19 @@ Two things the dev build does deliberately, both load-bearing:
   through `resolvePerModelConfigOptionSelection` for the same exemption, because a
   per-model option's ABSENCE from the snapshot is exactly what the probed model's
   lack of the control looks like and can never be the reason to reject another
-  model's turn. Runtime rejections remain in debug diagnostics; Codex/Claude mismatches
-  for model, reasoning effort, Fast, or Plan are not promoted to visible
-  `agent_warning` notices, while other rejected selections still are. Compatibility exception: Claude
-  Fable models omit the Fast mode option, so an explicit `fast=false` is skipped as
-  an already-effective no-op; `fast=true` must still be dispatched and retained in debug
-  diagnostics if rejected.
+  model's turn. Runtime rejections remain in debug diagnostics; what becomes a visible
+  `agent_warning` is DIVERGENCE — the state the agent publishes after applying the
+  turn's config contradicts what was requested. A rejection is not that signal in
+  either direction: Codex accepts `fast-mode` on a model with no fast speed tier and
+  then omits the option from its published state, so the state is the only evidence
+  Fast is off, while a rejected value that was already effective changed nothing and
+  must stay silent. Only where the agent published no config options at all (or for a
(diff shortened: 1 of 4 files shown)


zxch3n:
Superseded by #333, which contains these commits and supersedes their mechanism: the capability snapshot no longer rejects a run config at all, so the per-model exemption set this PR introduced has been deleted rather than maintained. Consolidating review on #333.

PR #333, opened 2026-09-03
feat: report ACP run-config divergence and let agents declare per-model capabilities

zxch3n:
## The bug this started from

`unknown ACP config option for the selected agent: fast-mode` — a Role that had
worked for weeks stopped dispatching after an upgrade.

The cause is a category error. `configOptions` from `session/new` is a
**snapshot of one model's state**, and every created session rewrites it, so it
describes whichever model ran last. Lody was reading it as an **agent-wide
capability catalog**. Codex publishes `fast-mode` only while the current model
has a fast speed tier, so a snapshot captured on a model without one became
evidence that the agent has no fast mode at all — and the request was rejected
before it ever reached the agent.

## The rule this PR establishes

**A snapshot reports; it does not reject.**

`validateTurnConfigOptionValues` now rejects only what no model could carry — a
value the option's own declared TYPE forbids. Everything else is dispatched and
reconciled against the state the agent publishes afterwards. A mismatch becomes
a visible `agent_warning`, not a blocked turn.

Agent Roles and frozen Operations behave exactly like ordinary preferences at
run time, differing only in warning wording. An upgrade must never turn a Role
that used to run into one that fails.

Inheritance is the one place that still drops values: a key carried forward from
an older turn is not a request anybody made, so a removed or renamed option must
not ride the Session lineage forever. Explicit `--config-option` and frozen
Operation requests go out unchanged.

## The one exception: permission

When the agent's own reported final state says the turn would run with **more**
permission than it asked for, the turn fails before `prompt`. By the time a
warning about that is readable the agent may already have edited files, so this
is the one divergence a notice cannot cover.

It fires only on a live contradiction — never from a snapshot, a stale cache, an
unranked third-party mode, an unconfirmed request, or a *narrower* outcome.

Three things make it correct:

- **Permission is applied last.** Claude rebuilds its permission modes on a model
  switch and downgrades to `default` when the new model lacks the current one, so
  a mode set before the model is silently widened by the model that follows.
- **All three permission shapes are checked** — the legacy `session/set_mode`
  selector, `category: 'mode'`, and `category: '_permission'` (Grok). Matching
  only the first two let a requested `ask` run as `always-approve`.
- **A successful `set_mode` is an acknowledgement, not a result.** It may fill a
  mode the agent's state does not report, never overwrite one.

The way out is explicit and per-turn: `acceptWiderPermissions` names the exact
`{ controlId, requestedModeId, effectiveModeId }` that was disclosed. It is a
list because differences are disclosed one stop at a time — two controls widening
at once produce two notices, and carrying only the newest would drop the previous
one and land the user back on it. Accumulating grants nothing extra, because
every entry is still matched exactly.

It must survive every rebuild on the way to the daemon, and must be dropped by
every derivation that mints a new `userTurnId`. The daemon enforces the drop, not
the client. `TurnScopedOverrides` is required rather than optional so a new send
route fails to compile instead of silently losing it.

## Letting the agent answer directly

The last piece removes the guessing. An adapter may now attach its own catalog to
`session/new`:

```jsonc
"_meta": { "lody": { "modelCapabilities": {
  "version": 1,
  "models": { "<modelId>": { "effortValues": ["low", "high"], "fastMode": true } }
} } }
```

Advisory in one direction only: it may describe a control the snapshot never
carried, and it never grants permission, authorizes a value, or rejects one. A
model it does not name is **unknown**, not unsupported — so it is read whole or
ignored whole, and gated by a TTL plus a `sourceVersion` match.

Producer side has landed:

- LodyAI/acp-extension-codex#31 → `9b4c961`
- LodyAI/acp-extension-claude#24 → `94223f1`

Both submodule pointers in this PR reference those merged commits on their mains.
Grok and Kimi have the same per-model shape and are the natural follow-ups;
DeepSeek Harness does not (its reasoning options are a fixed constant), which is
why it has no entry in `AGENT_PER_MODEL_BINDINGS`.

## Note on the lockfile

While this branch was catching up, `main` was red because `pnpm-lock.yaml`
recorded `@openai/codex` as `^0.153.3` while the codex submodule it pointed at
declared `^0.153.4`, so `pnpm install --frozen-lockfile` failed. The branch could
not install either, so it carried the one-line correction.

`main` has since been fixed independently and is green again. The two changes are
byte-identical, so this is now a no-op rather than a fix this PR delivers.

## Verification

`typecheck` · `lint` · `lint:i18n` · `check:public-boundary` ·
`check:platform-boundaries` clean.
Tests: shared 1018 · lody 2499 (4 skipped) · components 3159.

Each new mechanism was ablated — removed to confirm a test actually fails without
it. Three ablations passed silently on the first attempt and produced either a
real second bug (`runtimeConfigPatch.modeId` being overwritten by
`confirmedLegacyModeId`) or a strengthened test.

Supersedes #316 and #324.


timonwong:
Hi @zxch3n

Today I ran into an issue: “Fast mode” and “Reasoning Effort” disappeared for the Codex ACP provider. Will this PR fix that?

Snipaste_2026-09-23_17-33-45

image

PR #335, opened 2026-09-03
fix(cli): support custom ACP history synchronization

### LINKED ISSUE #187: [Bug] History sync from a custom ACP agent fails: no launch command configured
### Affected area

Agent runtime / ACP

### Installation method

Built from source

### Lody version or commit

966623d

### Operating system

macOS 26.4.1 arm64

### Agent or runtime

Any `cliType: custom` ACP agent (reproduced with a small read-only ACP server that implements `session/list` and `session/load`)

### What happened?

History sync from a custom ACP agent always fails with:

```
Custom ACP <agentType> has no launch command configured
```

A custom agent appears in the history-provider list — providers are built from every agent config on the machine, with no allowlist — so the UI offers it, but syncing from it can never succeed.

The cause is that `LocalProjectHistoryProvider` carries only `cliType` and `agentType`. That pair is enough for `builtin` and `registry` agents, whose executable resolves from a static table keyed by `agentType`. A `custom` agent's command is user-defined and lives on its agent config as `customAcp`, and nothing on the history path reads it, so `resolveCustomACPSetting` throws.

This makes the whole external-history feature unreachable for custom agents, which is the only way to import history from a tool that ships no ACP implementation of its own.

### What did you expect?

History sync to spawn the custom agent using the `customAcp` command recorded on its agent config, the same command the agent uses everywhere else.

### How can we reproduce it?

1. Register a custom ACP agent that implements `session/list` and `session/load`, and advertises `agentCapabilities.sessionCapabilities.list` and `loadSession`. A minimal stdio server is enough.
2. Add a local project.
3. Open project settings and sync history for that agent (or send `local-project/sync-history` over the project-control RPC with `provider: { cliType: 'custom', agentType: '<yours>' }`).
4. Observe the failure. The agent process is never spawned.

### How often does it happen?

Every time

### Relevant log output

```shell
project request parsed: type=local-project/sync-history
project request completed: type=local-project/sync-history ok=false error=execution_failed
Failed to sync history: Custom ACP <agentType> has no launch command configured
```

### Additional context

The provider reaches `resolveHistoryACPProcessLaunch` → `resolveACPProcessLaunchAsync` → `resolveACPSetting` → `resolveCustomACPSetting(agentType, input.customAcp)` with `customAcp` undefined, because it was never on the provider.

Worth noting for whoever picks this up: the history service is constructed in four independent places — `MessageHandler` plus three branches in the fleet dispatcher (sync, import, resolve-conflict) — and each passes the request's provider straight through. A fix applied to only one of them looks correct in review but leaves the path the CLI actually takes still broken; the symptom is identical, which makes it easy to conclude the fix did not work.

I have a working patch and am happy to open a PR if a maintainer agrees on the approach. Two shapes seem possible:

- resolve the launch spec on the daemon, looking the agent config up by `cliType:agentType` on the owning machine — leaves the control-plane schema unchanged and always uses the current command rather than one snapshotted when the request was built; or
- carry `customAcp` on the provider through the request — smaller, but it widens the wire contract and every construction site still has to be updated.

Either way a single owner for that resolution would prevent the next entry point from reintroducing it.

### Before submitting

- [X] I searched the existing issues and did not find a duplicate.
- [X] This report concerns an open-source component in this repository, not a hosted service, Web or mobile app, account, or billing issue.
- [X] This is not a security vulnerability; security reports follow the repository's security policy.
- [X] I removed credentials, private source, conversations, prompts, personal data, and other sensitive information.
- [X] If I plan to submit a pull request, I will wait for a Lody maintainer to explicitly agree on the scope and approach before implementation.

UniversePeak:
## Related issue

Closes #187

## Problem / pressure

The local-project history UI lists every agent configuration, including custom ACP agents, but the history sync launch path only received `cliType` and `agentType`. Builtin and registry agents can resolve their executables from static definitions; a custom agent's executable is stored in its machine-scoped `customAcp` configuration. As a result, every custom history sync failed before spawning the ACP process with `Custom ACP <agentType> has no launch command configured`.

## Summary

- Allow the local-project control guard to accept custom ACP history providers.
- Resolve the daemon-authoritative custom agent configuration for the requested machine and provider before history work starts.
- Thread the resolved `customAcp` launch spec through the shared history catalog client so session listing and replay use the same command and arguments as normal custom ACP sessions.
- Apply the resolved launch to sync, selected-session import, and conflict-resolution replay paths without widening the control-plane provider wire contract.
- Add regression coverage for custom launch resolution and custom history control requests.

## Before / after

| Before | After |
| ------ | ----- |
| Custom ACP providers appeared in local-project history settings but were rejected by the remote control guard or reached launch resolution without a command. | Custom providers pass the control boundary, and the owning daemon resolves their current machine-scoped `customAcp` command before spawning history ACP processes. |
| History listing/import/conflict resolution could never start a custom ACP process. | All history entry points share the resolved custom launch while builtin and registry behavior remains unchanged. |

## Test plan

- `corepack pnpm --filter acp-extension-core build` — passed, required local ACP package entry was built.
- `corepack pnpm --filter acp-extension-claude build` — passed, required local workspace entry was built.
- `corepack pnpm --filter lody typecheck` — passed.
- `corepack pnpm --filter @lody/shared typecheck` — passed.
- `corepack pnpm --filter lody exec vitest run tests/history-session-catalog-client.test.ts -t 'custom ACP'` — passed, 1 test.
- `corepack pnpm --filter @lody/shared exec vitest run tests/local-project-control.test.ts` — passed, 29 tests.
- `corepack pnpm --filter lody exec vitest run tests/local-project-history-sync-service.test.ts` — passed, 30 tests.
- Focused type-aware oxlint on changed CLI/shared files — 0 errors.
- Prettier checks on all changed files and `git diff --check` — passed.
- Full `@lody/shared` suite — 1,012 passed, 3 unrelated baseline failures (one local CLI host-lease assertion and two source-guard tests timed out); no changed history-control test failed.

- **Review focus:** Check `LocalProjectHistorySyncService.resolveLaunchProvider`, the three history operation paths, and `history-session-catalog-client.ts` to ensure only daemon-authoritative machine config supplies the custom command.
- **Decisions to challenge:** Confirm that resolving by machine plus `cliType:agentType` is the right identity boundary and that keeping `customAcp` out of the control-plane wire schema avoids trusting caller-supplied executable paths.
- **Plausible failures / evidence gaps:** The full shared suite retains three unrelated baseline failures; live spawning of a user-provided ACP server was not run, while the launch resolver regression verifies exact command and argv propagation.

### Authoring context

- **User goal / directives:** Make local-project history synchronization work for custom ACP agents using the command already stored in their agent configuration.
- **Constraints / non-goals:** Keep builtin and registry launches unchanged, do not accept executable paths from the RPC caller, do not duplicate the custom command in session metadata, and do not broaden the change beyond local-project history operations.
- **Risk-bearing decisions:** The daemon reads the merged config for the target machine and provider, then passes that trusted launch spec through list/replay; a missing config or command retains the existing clear failure instead of falling back.
- **Destructive or irreversible behavior:** None; this changes process-launch selection and adds tests only, with no persistence migration or user-data rewrite.
- **Deliberately not done or tested:** No live custom ACP server was launched and no full desktop build was run; focused type, lint, history-service, control-schema, and launch-resolution checks cover the changed paths.
- **Unknowns / confidence:** High confidence in the control and launch plumbing; the remaining uncertainty is compatibility with varied third-party custom ACP servers, which CI and maintainer review can exercise.

### DIFF
--- a/apps/cli/src/lib/history-session-catalog-client.ts
+++ b/apps/cli/src/lib/history-session-catalog-client.ts
@@ -24,6 +24,7 @@ import {
   type ACPSessionId,
   getLocalProjectHistoryProviderKey,
   type LocalProjectHistoryProvider,
+  type CustomAcpLaunchSpec,
 } from '@lody/shared';
 import { LODY_EXTENSION_METHODS } from 'acp-extension-core';
 
@@ -126,11 +127,19 @@ type ResolvedHistoryACPProcessLaunch = Omit<ResolvedACPProcessLaunch, 'env'> & {
   env: NodeJS.ProcessEnv;
 };
 
+export type HistoryLaunchProvider = LocalProjectHistoryProvider & {
+  customAcp?: CustomAcpLaunchSpec;
+};
+
 export async function resolveHistoryACPProcessLaunch(args: {
-  provider: LocalProjectHistoryProvider;
+  provider: HistoryLaunchProvider;
   env?: NodeJS.ProcessEnv;
 }): Promise<ResolvedHistoryACPProcessLaunch> {
-  const launch = await resolveACPProcessLaunchAsync(args.provider);
+  const launch = await resolveACPProcessLaunchAsync({
+    cliType: args.provider.cliType,
+    agentType: args.provider.agentType,
+    customAcp: args.provider.customAcp,
+  });
   return {
     ...launch,
     env: mergeACPProcessEnv(launch, args.env ?? process.env),
@@ -138,7 +147,7 @@ export async function resolveHistoryACPProcessLaunch(args: {
 }
 
 async function createHistoryAcpConnection(args: {
-  provider: LocalProjectHistoryProvider;
+  provider: HistoryLaunchProvider;
   workdir: string;
   logger: Logger;
 }): Promise<HistoryAcpConnection> {
@@ -327,7 +336,7 @@ export function dedupeHistorySessionsById(sessions: SessionInfo[]): SessionInfo[
 }
 
 export async function listHistorySessionsForLocalProject(args: {
-  provider: LocalProjectHistoryProvider;
+  provider: HistoryLaunchProvider;
   rootPath: string;
   logger: Logger;
   requiredSessionIds?: readonly string[];
@@ -376,7 +385,7 @@ export async function listHistorySessionsForLocalProject(args: {
 }
 
 export async function loadHistorySessionReplay(args: {
-  provider: LocalProjectHistoryProvider;
+  provider: HistoryLaunchProvider;
   rootPath: string;
   acpSessionId: ACPSessionId;
   logger: Logger;
--- a/apps/cli/src/lib/history-session-catalog-client.ts
+++ b/apps/cli/src/lib/history-session-catalog-client.ts
@@ -25,6 +25,7 @@ import {
   getLocalProjectHistoryProviderKey,
   type LocalProjectHistoryProvider,
   type CustomAcpLaunchSpec,
+  type BuiltinRuntimeOverrides,
 } from '@lody/shared';
 import { LODY_EXTENSION_METHODS } from 'acp-extension-core';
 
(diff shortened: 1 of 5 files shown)


UniversePeak:
Addressed the P1 review in commit `87f89b4`.

History sync now resolves the selected machine's authoritative `AgentConfigMeta` and propagates its `env` alongside `customAcp` for session listing, import, and conflict replay. The caller cannot inject an executable path; the launch still comes from the resolved machine configuration. Added a regression test covering config environment precedence for API key, base URL, and PATH. Targeted history-sync tests, CLI/shared typechecks, changed-file oxlint, Prettier, and `git diff --check` pass; the remaining full-suite failures are pre-existing Windows/path and environment/submodule issues.

PR #360, opened 2026-09-03
fix(mcp): preserve invoking turn identity

wibus-wee:
## Summary

- use the active execution runtime as the authoritative source of MCP invocation identity; never reconstruct identity from persisted history
- route Session ownership, member-scoped Machine/GitHub access, and downstream Turn attribution through that runtime requester
- freeze source-Turn and requester provenance across Operation retries, recovery, and completion continuations; reject cross-Turn operation id reuse

## Validation

- `pnpm --filter lody typecheck`
- `pnpm --filter lody exec vitest run src/mcp/lody-mcp-server.test.ts src/mcp/lody-mcp-server-chat-sync.test.ts src/orchestration/operation-store.test.ts src/orchestration/operation-coordinator.test.ts`
- `pnpm --filter lody exec vitest run src/commands/session.test.ts -t "delegated requester identity"`
- scoped Prettier check
- `pnpm check:public-boundary`
- scoped type-aware Oxlint (0 errors)

Full `session.test.ts` was also attempted. Three existing Git fixture tests fail before assertions because this machine enables global GPG signing without the corresponding private key; all identity-related tests pass.

The HTTP MCP host continues to use one daemon bearer token with unsigned session/workspace/machine context headers. This is currently acceptable under the existing trust model where agent runtimes and subprocesses under the same OS user are trusted. The active session header now participates in requester/principal selection, so an untrusted coding agent would require a session-bound capability or signed context in a future hardening change.


wibus-wee:
Non-blocking architectural/security note: the HTTP MCP host still uses one daemon-wide bearer token, while session/workspace/machine identity is supplied by unsigned headers. Since `x-lody-mcp-session-id` now feeds active invocation lookup and therefore requester selection, the existing host comment that “header-supplied ids grant nothing the agent could not already reach” is no longer literally accurate.

This remains acceptable if the intended threat model is that agent runtimes and subprocesses under the same OS user are trusted. If coding agents become untrusted in the future, this boundary needs a session-bound capability or signed context rather than relying on a shared bearer plus unsigned session header. This does not block #360 identity correctness.

wibus-wee:
Implementation LGTM after syncing origin/main at 24077862. The active execution runtime is authoritative for invocation identity; persisted history is not used for identity reconstruction. New CI is green on the updated base/head. The HTTP MCP shared-bearer and unsigned-context-header concern remains non-blocking under the documented same-OS-user trusted-runtime assumption.

PR #367, opened 2026-09-04
fix(acp): route Factory Droid through direct ACP

### LINKED ISSUE #366: [Bug] Factory Droid ACP daemon fails with stdio MCP servers
### Affected area

Agent runtime / ACP

### Installation method

npm / npx CLI

### Lody version or commit

Lody source commit `ec3bef5` (the commit before the proposed fix)

### Operating system

Linux x86_64

### Agent or runtime

Factory Droid CLI `0.211.0`, using ACP with one synthetic stdio MCP server

### What happened?

When Lody starts Factory Droid through `acp-daemon` while a stdio MCP server is selected, `session/new` fails with `-32603 Internal error`. The Droid child exits with code 1, so the ACP session cannot start with MCP enabled.

The failure only occurs when the ACP request includes a non-empty MCP server list. The same request succeeds when Factory Droid is started through direct `acp`.

### What did you expect?

`session/new` should succeed and the selected MCP server should be available to the Droid session. The daemon should either pass the MCP configuration through a supported protocol or avoid injecting unsupported command-line arguments into its child process.

### How can we reproduce it?

1. Install Factory Droid `0.211.0` through `npx`.
2. Build or run Lody from source before the proposed registry change, with Factory Droid configured to use `droid exec --output-format acp-daemon`.
3. Configure one synthetic stdio MCP server that implements the MCP initialize exchange; do not use credentials or private services.
4. Start a new Factory Droid ACP session so Lody sends the MCP server in the `session/new.mcpServers` request.
5. Observe `session/new` returning `-32603 Internal error` and the Droid child exiting with code 1.
6. Change only the Factory Droid launch mode from `acp-daemon` to direct `acp` and repeat the request. The session starts successfully with the same MCP server configuration.

### How often does it happen?

Every time

### Relevant log output

```text
session/new -> -32603 Internal error
Factory Droid child -> exit code 1
child stderr: error: unknown option '--mcp-servers'
```

### Additional context

Tracing the failing daemon launch shows that its parent starts an inner Droid process with an injected argument shaped like `--mcp-servers <base64-json>`. The inner `droid exec --output-format acp` command rejects that argument as unknown and exits; the daemon wraps the child failure as `-32603`.

An ablation across Factory Droid `0.208.2` and `0.211.0` shows the same result: both versions fail with `acp-daemon` and succeed with direct `acp`. This indicates a mode-specific Droid daemon/child argument contract issue rather than a Lody version-specific or authentication issue. Lody's ACP request already carries MCP configuration through `session/new.mcpServers`; the direct-ACP registry workaround is included in the accompanying pull request.

### Before submitting

specter119:
## Related issue

Closes #366

## Problem / pressure

Factory Droid's `acp-daemon` mode fails during ACP `session/new` whenever Lody includes a stdio MCP server. The daemon injects an unsupported `--mcp-servers` argument into its inner Droid process, which exits with code 1 and is surfaced as `-32603 Internal error`. This blocks the normal MCP-enabled Factory Droid workflow.

## Summary

- Exclude the broken remote Factory Droid registry entry.
- Add a pinned `droid@0.211.0` npx registry entry that launches direct `acp` with auto-update disabled.
- Add a regression test covering the pinned version, distribution, launch arguments, and capability source version.

## Before / after

| Before | After |
| ------ | ----- |
| Factory Droid used `droid@0.208.2 exec --output-format acp-daemon`; with a non-empty `session/new.mcpServers`, the daemon launched a child with unsupported `--mcp-servers` arguments and ACP returned `-32603`. | Lody uses pinned `droid@0.211.0 exec --output-format acp`; MCP configuration remains in the ACP `session/new.mcpServers` request and session initialization succeeds. |

## Test plan

- Factory Droid ACP ablation with a synthetic stdio MCP server: `acp-daemon` failed and direct `acp` succeeded on both Droid `0.208.2` and `0.211.0`.
- `apps/cli/tests/agent-setting.test.ts`: 30 tests passed.
- `tsgo --noEmit`: passed.
- `node scripts/check-public-boundary.mjs`: passed.
- `git diff --check`: passed.
- PR body validation with `.github/scripts/check-pr-body.mjs`: passed.

- **Review focus:** Check the generated and source ACP registry entries plus the agent-setting regression test, especially direct-ACP arguments, version pinning, and MCP capability behavior.
- **Decisions to challenge:** Confirm that bypassing `acp-daemon` is the correct compatibility boundary and that keeping MCP in `session/new.mcpServers` preserves the intended protocol contract.
- **Plausible failures / evidence gaps:** Future Droid releases may change ACP flags, and CI does not run a live Droid/MCP integration; the reported failure was reproduced locally with synthetic stdio MCP input.

### Authoring context

- **User goal / directives:** Restore reliable Factory Droid ACP sessions in Lody when stdio MCP servers are selected, and provide an upstream contribution that documents the failure.
- **Constraints / non-goals:** Keep the fix within Lody's public ACP registry and tests; do not patch Factory Droid, add private dependencies, or include private transcripts, credentials, or service data.
- **Risk-bearing decisions:** Pin Droid `0.211.0`, disable its auto-update path, and select direct `acp` so Lody does not enter the daemon's unsupported child-argument bridge.
- **Destructive or irreversible behavior:** None. The change affects agent process selection and registry metadata only; it performs no data migration, deletion, overwrite, or runtime rollback.
- **Deliberately not done or tested:** Factory Droid itself was not patched; no live authenticated provider or private MCP service was used. The integration reproduction used a synthetic stdio MCP server.
- **Unknowns / confidence:** Confidence is high because child launch tracing and a version/mode ablation isolated the unsupported `--mcp-servers` argument. The remaining uncertainty is future compatibility if Droid changes its ACP CLI contract.


scripts/generate-acp-registry.mjs
```diff
@@ -9,10 +9,12 @@ const EXCLUDED_REMOTE_REGISTRY_AGENT_IDS = new Set([
   'claude-acp',
   'claude-p',
   'codex-acp',
+  'factory-droid',
   'grok-build',
 ]);
 const OFFICIAL_NPM_REGISTRY = 'https://registry.npmjs.org/';
 const INTERACTIVE_CLAUDE_ACP_VERSION = '0.1.5';
+const FACTORY_DROID_ACP_VERSION = '0.211.0';
```
Leeeon233: I donot wanna fix the version of Droid. Maybe you should fetch the registry, parse it and replace the command.

Leeeon233 (changes requested):

scripts/generate-acp-registry.mjs
specter119: @Leeeon233 recovered the dynamic version.

specter119:
@Leeeon233 plaese review

Leeeon233:
hi @specter119 please resolve the conflicts

PR #382, opened 2026-09-04
fix(cli): resolve archive worktrees from machine Flock

### LINKED ISSUE #377: [Bug] Archived sessions cannot be cleared from worktree
### Affected area

Desktop app

### Installation method

Desktop release

### Lody version or commit

0.90.1

### Operating system

MacOS 26.6

### Agent or runtime

PI 0.84.4

### What happened?

After archiving the session, it is not possible to clean up the working tree and execute the cleanup script

### What did you expect?

After archiving the session, clean up the worktree and execute the cleanup script

### How can we reproduce it?

1. Archive Session
2. The working tree directory corresponding to the session is still present, and the cleanup script has not been executed.

### How often does it happen?

Every time

### Relevant log output

```shell

```

### Additional context

_No response_

### Before submitting

UniversePeak:
## Related issue

Closes #377

## Problem / pressure

Archiving a local-project session did not remove its worktree or run the configured cleanup script when the project metadata was stored only in the machine Flock document. The archive path resolved local projects from legacy machine metadata but did not load the Flock rows that are authoritative for current local projects.

## Summary

Load the machine Flock command snapshot's `localProject` rows while processing archive requests and pass them to worktree cleanup target resolution. Add a regression that creates a real local Git worktree with project metadata available only through the machine Flock and verifies archiving removes that worktree.

## Visual explanation

```mermaid
flowchart TD
  A[Queued archive request] --> B[Wait for initial Machine Flock sync]
  B --> C[Read authoritative Machine Flock snapshot]
  C --> D[Resolve localProject root path]
  D --> E[Run archive cleanup / remove Lody worktree]
  E --> F[Acknowledge archive request]
```

The archive request stays queued until the Machine Flock watcher has completed its initial cloud sync; cleanup and acknowledgement then use the same authoritative snapshot.

## Before / after

| Before | After |
| ------ | ----- |
| Archive cleanup could not resolve a local project's root path when it existed only in Machine Flock, so cleanup and worktree removal were skipped. | Archive cleanup resolves the root path from the same Machine Flock local-project catalog used by deletion, then removes the archived worktree. |

## Test plan

- `corepack pnpm install --frozen-lockfile --ignore-scripts` — passed.
- `corepack pnpm --dir packages/acp-extension-core build`, plus Claude, Codex, Grok, and DSH adapter builds — passed; required for the CLI test imports.
- `corepack pnpm --dir apps/cli exec vitest run tests/message-handler-terminal-cleanup.test.ts` — 13 tests passed, including the new archive regression.
- `corepack pnpm --filter lody typecheck` — passed.
- `corepack pnpm lint:fast` — 0 errors (465 existing warnings).
- `corepack pnpm check:public-boundary` — passed.
- Prettier check on changed files and `git diff --check` — passed.

- **Review focus:** Check that archive processing reads `localProject` rows from Machine Flock and passes them to `resolveWorktreeCleanupTarget`, matching the existing permanent-delete path.
- **Decisions to challenge:** Confirm that one shared command-row reader should include the local-project catalog while preserving the existing legacy metadata fallback.
- **Plausible failures / evidence gaps:** The regression exercises native worktree removal; a configured user cleanup script is not needed to prove the previously skipped target-resolution branch.

### Authoring context

- **User goal / directives:** Make archived local-project sessions clean up their worktree instead of leaving the directory and branch registered.
- **Constraints / non-goals:** Keep the change limited to archive target resolution; do not alter dirty-worktree safety, cleanup script semantics, or the deletion path.
- **Risk-bearing decisions:** Read the authoritative local-project catalog from Machine Flock and retain legacy machine metadata as a fallback for older clients.
- **Destructive or irreversible behavior:** Archive now removes the Lody-owned worktree through the existing forced archive path; the preserved branch behavior and configured cleanup hook remain unchanged.
- **Deliberately not done or tested:** No live desktop release was run; the focused MessageHandler suite uses synthetic session metadata and a real temporary Git repository.
- **Unknowns / confidence:** High confidence in the missing-row diagnosis and focused regression; broader Electron UI behavior remains covered by existing integration paths.

### DIFF
--- a/apps/cli/src/lib/message-handler.ts
+++ b/apps/cli/src/lib/message-handler.ts
@@ -3810,7 +3810,12 @@ export class MessageHandler {
       this.getMachineFlockDocIdForMachine()
     );
     return readMachineFlockRowsFromFlock(handle.flock, {
-      families: ['archiveSessionCommand', 'deleteSessionCommand', 'deleteLocalProjectCommand'],
+      families: [
+        'archiveSessionCommand',
+        'deleteSessionCommand',
+        'deleteLocalProjectCommand',
+        'localProject',
+      ],
     });
   }
 
@@ -3968,9 +3973,10 @@ export class MessageHandler {
     this.logger.debug(`[${sessionId}] Preview tunnel closed for archive`);
 
     const sessionRoomId = getSessionRoomId(sessionId);
-    const [sessionMetaDoc, archiveMachineMetaDoc] = await Promise.all([
+    const [sessionMetaDoc, archiveMachineMetaDoc, archiveMachineFlockRows] = await Promise.all([
       this.workspaceDocument.repo.getDocMeta(sessionRoomId),
       this.workspaceDocument.repo.getDocMeta(getMachineRoomId(this.machineId)),
+      this.tryReadMachineFlockCommandRows(),
     ]);
     const sessionMeta = sessionMetaDoc?.meta as SessionMeta | undefined;
     const archiveMachineMeta = archiveMachineMetaDoc?.meta as MachineLegacyMetaFields | undefined;
@@ -3986,6 +3992,7 @@ export class MessageHandler {
       const cleanupTarget = this.resolveWorktreeCleanupTarget({
         sessionMeta,
         machineMeta: archiveMachineMeta,
+        machineFlockRows: archiveMachineFlockRows,
       });
       if (cleanupTarget) {
         const worktreeManager = getWorktreeManager(this.buildWorktreeManagerConfig(cleanupTarget));
--- a/apps/cli/src/lib/message-handler.ts
+++ b/apps/cli/src/lib/message-handler.ts
@@ -3925,6 +3925,14 @@ export class MessageHandler {
   }
 
   private async processArchiveRequests(): Promise<void> {
+    // Archive cleanup may need a local-project root that exists only in Machine
+    // Flock. In cloud mode, wait until its command watcher has completed the
+    // initial remote sync; a successful read before then can still be stale.
+    if (this.cloudPort.kind !== 'local' && !this.machineFlockCommandWatcher.isReady) {
+      this.logger.debug('[archive] Waiting for authoritative Machine Flock before processing');
+      return;
+    }
+
     let snapshot: MachineCommandSnapshot;
     try {
       snapshot = await this.readMachineCommandSnapshot({ strictFlock: true });
--- a/apps/cli/src/lib/message-handler.ts
+++ b/apps/cli/src/lib/message-handler.ts
@@ -3833,11 +3833,16 @@ export class MessageHandler {
     }
   }
 
-  private async readMachineCommandSnapshot(): Promise<MachineCommandSnapshot> {
+  private async readMachineCommandSnapshot(options?: {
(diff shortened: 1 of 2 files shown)


wibus-wee (changes requested):
I think this still has one reliability gap.

`processArchiveRequests()` already reads the Machine Flock successfully, but `archiveSessionResources()` reads it again. That second read is best-effort: if it fails, it returns an empty result.

For local projects whose `rootPath` only exists in Machine Flock, that means a temporary read failure can make us skip the cleanup entirely, but still mark the archive as successful and remove the archive request. At that point we won't retry, and the worktree can be left behind permanently.

Could we reuse the Machine Flock snapshot we already read in `processArchiveRequests()`, or fail the archive when the cleanup target cannot be resolved because that read failed?

It would also be good to add a regression test for this failure case, not only the successful Flock-read path.

UniversePeak:
Addressed in bcab8ae4. `processArchiveRequests()` now treats the Machine Flock read as authoritative for archive processing, reuses that snapshot in `archiveSessionResources()`, and leaves the archive request queued when the read fails so a later watcher/retry can recover. I also added a regression covering the transient read failure. Local verification passes: 14 focused tests, CLI typecheck, oxlint, Prettier, and `git diff --check`.

UniversePeak:
Updated the fix in c731b12c (the branch was amended after the initial push). The archive path still reuses the authoritative Machine Flock snapshot and keeps the request queued when that snapshot cannot be read; the regression now exercises the failure guard. Re-ran the focused suite (14 passed), CLI typecheck, oxlint, Prettier, and `git diff --check`. The two required CI jobs for the new head are awaiting GitHub approval/action, while the prior head’s Static checks and Tests were green.

wibus-wee:
Thanks, but i think there is still one race here.

`strictFlock: true` only means “fail if reading Flock throws”. It does not mean the Flock data is already synced and authoritative.

On startup, `setupArchiveWatcher()` calls `processArchiveRequests()` before the Machine Flock watcher finishes its first remote sync. At that point the legacy `needToArchiveSessions` queue may already contain the archive request, while the `localProject` row has not arrived yet.

In that case the Flock read succeeds, but `rootPath` is still missing. We then get no cleanup target, skip the cleanup/worktree removal, and still acknowledge the archive request. That leaves the same bug permanently.

Could we avoid processing archive requests until the Machine Flock watcher is ready/authoritative, or otherwise retry when the local project metadata may still be unsynced?

One meta suggestion as well: this kind of lifecycle / durability fix has a lot of subtle race conditions, and the earlier iterations missed a couple of failure modes that only show up when you challenge the happy path pretty aggressively.

If possible, I’d avoid relying on Luna alone for the implementation/review pass here. Please use a stronger model / higher reasoning setting for the final self-review before considering the fix complete.

UniversePeak:
Updated in 03e2a9f0. Archive processing now waits for the Machine Flock command watcher to finish its initial remote sync in cloud mode, so a queued legacy archive request cannot be acknowledged from a successful but stale snapshot that lacks the local-project root path. The new regression keeps a real worktree and request intact before authority, then verifies the authoritative retry removes the worktree and clears the request. Focused cleanup tests (15), CLI typecheck, Prettier, and `git diff --check` pass.

PR #402, opened 2026-09-04
perf(shared): buffer fragmented local data-plane lines

slashdevcorpse:
## Related issue

Refs #240

## Problem / pressure

The local data-plane JSON line splitter concatenates and searches the entire pending line on every socket chunk. Large fragmented frames repeatedly scan and flatten the same prefix. Both Electron and the CLI use this shared splitter.

## Summary

Keep decoded fragments and their character count until a newline arrives, then join once. Complete lines within one chunk still use direct slices. Preserve the decoder, trimming, overflow recovery, and unread input after a callback exception.

This is the isolated chunk-buffering slice of #240. The current implementation lives in `packages/shared/src/local-loro-data-plane.ts`, consumed by the Electron relay and CLI server. It does not alter RPC transport, renderer rows, usage queries, wire formats, or dependencies. The remaining issue scope stays open.

## Before / after

Synthetic splitter benchmark on Windows, Node 24.15.0, pnpm 10.20.0, Vitest 3.2.4. Values are mean milliseconds per operation, with fixture creation and full-output verification outside timing.

| Input | Before | After |
| --- | ---: | ---: |
| 1.1 MB ASCII payload, 64-byte chunks | 1795.11 | 2.3141 |
| 1.1 MB ASCII payload, 4096-byte chunks | 27.9515 | 0.2593 |
| 1.1 MB ASCII payload, 65536-byte chunks | 2.0961 | 0.2209 |
| 10,000 short frames, 64 KiB batches | 0.2764 | 0.1494 |

The baseline is `d0e5059`. The same benchmark ran before and after the source change. The slowest baseline had three samples; these are microbenchmark means, not application p95 measurements. No renderer heap, FPS, thread-open, or compression improvement is claimed.

## Test plan

- `corepack pnpm --filter @lody/shared exec vitest bench --run local-loro-data-plane-splitter` passed before and after the change. The committed benchmark verifies complete output and can be rerun without timing thresholds.
- Seven splitter tests passed, including fragmented 1.1 MB frames, multibyte boundaries, trimming and cap equality, overflow recovery, and callback-exception continuation.
- `corepack pnpm --filter @lody/shared typecheck` passed.
- `corepack pnpm --filter lody test local-loro-data-plane-bug-repro --maxWorkers=2` passed the real socket overflow/recovery regression.
- `node --test apps/electron/src/main/services/loro-data-plane-relay.test.mjs` passed.
- Changed-file Prettier, `git diff --check`, and `node scripts/check-public-boundary.mjs` passed.
- `corepack pnpm check` did not pass. Lint reported zero errors; the shared suite passed 1,010 tests but failed the daemon shutdown callback assertion in `local-cli-host-lease.test.ts:133`. The root typecheck command also reported no projects matching its recursive filters, so shared typecheck was run directly.
- The separate Electron suite also failed Sparkle packaging/path tests on Windows. Both failures reproduce on unchanged `d0e5059`. The host-lease test and implementation match baseline byte-for-byte, but its baseline run could not resolve Vitest in the dependency-free clone. These areas are outside this PR. Full desktop interaction and application-level profiling were not run.

- **Review focus:** Check splitter fragment lifetime, newline order, and recovery after overflow or callback exceptions in `packages/shared/src/local-loro-data-plane.ts`.
- **Decisions to challenge:** Retain the existing character-count cap and trim-before-cap behavior for complete lines while eliminating repeated prefix scans.
- **Plausible failures / evidence gaps:** Tiny fragments retain array entries until completion; benchmark results isolate framing and do not prove desktop heap or latency targets.

### Authoring context

- **User goal / directives:** Deliver a focused performance contribution related to the long-session issue with reproducible verification.
- **Constraints / non-goals:** No protocol, schema, dependency, UI, settings, or hosted-service changes; keep the broader issue open.
- **Risk-bearing decisions:** Use decoded string fragments so the existing stateful UTF-8 decoder remains authoritative; retain unread suffixes when callbacks throw.
- **Destructive or irreversible behavior:** No storage deletion, migration, persistent-data rewrite, or deployment. Reverting the commit restores the previous splitter.
- **Deliberately not done or tested:** Renderer optimization and application p95, heap, FPS, and transfer-budget work remain outside this slice; no full desktop session benchmark was run.
- **Unknowns / confidence:** Focused framing tests and the socket regression pass; independent source review found no blocking defects, but broader Windows suites are not green.

### DIFF
--- a/packages/shared/src/local-loro-data-plane.ts
+++ b/packages/shared/src/local-loro-data-plane.ts
@@ -578,10 +578,16 @@ export function createJsonLineSplitter(options: {
   onOverflow?: () => void;
 }): (chunk: string | Uint8Array) => void {
   const decodeChunk = createUtf8StreamDecoder();
-  let buffer = '';
+  let parts: string[] = [];
+  let bufferedLength = 0;
+  let retryChunk = '';
   let discardingOversizedLine = false;
   return (input: string | Uint8Array) => {
     let chunk = typeof input === 'string' ? input : decodeChunk(input);
+    if (retryChunk) {
+      chunk = retryChunk + chunk;
+      retryChunk = '';
+    }
     if (discardingOversizedLine) {
       const newlineIndex = chunk.indexOf('\n');
       if (newlineIndex < 0) {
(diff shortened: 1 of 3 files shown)


Leeeon233:
Huge!

PR #409, opened 2026-09-05
fix(cli): keep a finished turn's endedAt across teardown finalize

### LINKED ISSUE #260: [Bug] Completed turn duration grows after the app is quit and reopened
### Affected area

Session / workspace / local sync

### Installation method

Built from source

### Lody version or commit

`0.76.0` (`1cc6553`)

### Operating system

Linux 6.8.0-138-generic x64

### Agent or runtime

Codex `0.148.0` (the managed runtime Lody downloaded), model `5.6-Sol`

### What happened?

A finished short reply showed **Worked for 8s**. I fully quit the Lody desktop app, opened the same session again, and that same reply now showed **Worked for 1m 02s**. The number got larger and did not go back.

Leaving the session open and idle did **not** change it. It only jumped after the process exited and the session was loaded again.

I saw the same rewrite on three finished turns. The extra time is about as long as the wall clock from "the reply was already on screen" to "I killed the process" — not the time the model spent generating.

| Turn | Prompt (verbatim) | On screen when it finished | After quit + reopen |
| --- | --- | --- | --- |
| ping3 | `Reply with exactly the single word ping3.` | **8s** | **1m 02s** |
| pong | `Reply with exactly the single word pong and nothing else.` | **6s** | **1m 04s** |
| pong2 | `Reply with exactly the single word pong2 and nothing else.` | **8s** | **4m 22s** |

ping3 is the cleanest row: I sent it at 14:21:07, the answer was on screen at 14:21:17 (~10s wall clock, label 8s), waited ~40s idle (still 8s), then quit. After reopen it was 1m 02s. pong2 sat in a running app for several more minutes before quit, and grew the most.

### What did you expect?

A turn that already displayed 8s should still display 8s after the app is quit and the same session is opened again.

### How can we reproduce it?

1. Build and run the OSS desktop from `1cc6553` (`pnpm start:local`). Use a Codex session on a model that will answer quickly.
2. Send exactly:

   `Reply with exactly the single word ping3.`

3. Wait until the assistant message is on screen. Read **Worked for …** under that reply (I got **8s**).
4. Do not send anything else. Leave the session open about 40 seconds. Confirm the label is still **8s**.
5. Fully quit the Lody desktop process (not only switching chats or hiding the window).
6. Start Lody again, open the same session, do not send a new message.
7. The same ping3 reply now shows a larger **Worked for** (I got **1m 02s**). Opening it a third time keeps the larger number.

### How often does it happen?

Every time

### Relevant log output

No crash or error log. The duration label on the finished turn is the symptom.

### Additional context

What I saw in the UI is only the label change above, including that idle does not grow the number.

The next paragraph is from reading the tree at `1cc6553`, not from extra UI steps.

The desktop formats `endedAt - timestamp` via `resolveSessionHistoryDurationMs` (`packages/components/src/lib/session-history-duration.ts`). While the turn is finishing, that difference is the 8s I saw. On session `exit` / `terminated`, `apps/cli/src/lib/message-handler.ts` calls `finalizeACPState(sessionId)` with no `turnId` (`:3722` and `:3742`). That function always does `const endedAt = Date.now()` (`:5729`) and writes it onto the last assistant entry (`:5750–5752`) even when that entry already has `endedAt`. Reloading the session then shows shutdown time minus start time.

A teardown path that does not overwrite `endedAt` on a turn that already completed would match what the UI showed before quit. You will know better which write should be skipped.

Screenshots of ping3 at 8s (right after the reply, and again after idle) and at 1m 02s after reopen are attached.

ping3 刚答完：Worked for 8s

同一条再闲置约 40s：仍是 8s

整进程退出再打开同一会话：ping3 变成 Worked for 1m 02s

### Before submitting

audichuang:
## Related issue

Closes #260

## Problem / pressure

A finished assistant turn's "Worked for X" grew by however long the app stayed
open after that turn ended. It does not tick up while the app is running: the
stored `endedAt` is rewritten exactly once, at teardown, so the displayed
number jumps the next time the session is loaded.

The renderer derives the duration from `endedAt - timestamp`
(`packages/components/src/lib/session-history-duration.ts`
`resolveSessionHistoryDurationMs`), so the inflation comes from `endedAt` being
rewritten after the turn is over.

`finalizeACPState` (`apps/cli/src/lib/message-handler.ts`) has a no-turnId
overload. Without a `turnId` its history loop matches the last assistant entry
regardless of whether that entry already finished, and unconditionally writes
`endedAt = Date.now()`. On app close, `sessionManager.cleanUp()` raises
`exit`/`terminated` for every live session, and both handlers call that overload
after checking only `this.store.has(sessionId)` — no pending-work check. A
session whose turn ended an hour earlier is re-stamped with the close-time clock.

(`flushAllACPUpdates()` is guarded by `hasPendingTurnWork` and skips an idle
turn with an empty buffer, so it is not the unconditional trigger; the two
teardown handlers are.)

`apps/cli/src/session/AGENTS.md` already documents the intended contract — the
no-turnId overload "stamps `finished=true`/`endedAt` on the **in-progress**
entry". The implementation did not honour that.

## Summary

- New `apps/cli/src/lib/assistant-turn-finalize.ts` exports
  `markAssistantTurnFinished(history, { turnId, endedAt, permissionWaitMs })`:
  the finalize history loop, verbatim, plus one guard — an entry already marked
  `finished` is left untouched.
- `finalizeACPState` now calls it instead of inlining the loop, so the guard
  covers every no-turnId call site, not just the teardown pair above.
- The guard tests `finished === true` rather than filling in a missing
  `endedAt`: `createAssistantImageGroupEntry` and `createAssistantFileEntry`
  publish assistant entries with `finished: true` and no `endedAt`, so an
  `endedAt ??=` guard would still stamp close time on them. Leaving `endedAt`
  absent makes `resolveSessionHistoryDurationMs` return `null` and the row shows
  no duration, which is the honest answer.
- A turn interrupted mid-flight is not `finished`, so its teardown stamp still
  lands; resume still clears the footprint through `writeAssistantEntryForTurn`'s
  reopen branch. Behaviour there is unchanged.
- `apps/cli/src/session/AGENTS.md` records that the overload is a no-op on an
  already-finished entry.

## Before / after

| Before | After |
| ------ | ----- |
| Every no-turnId finalize rewrote the last assistant entry's `endedAt` to `Date.now()`, so closing the app inflated a finished turn's "Worked for X" by the time the app had been open. | The first terminal stamp is final; a later teardown finalize for an already-finished turn writes nothing and the displayed duration stays put. |
| The finalize history loop was inline in `finalizeACPState`, with nothing asserting that a second finalize is harmless. | The rule is one named function, covered directly and through `finalizeACPState` itself. |

## Test plan

- New `apps/cli/tests/message-handler-turn-duration.test.ts` drives the real
  `finalizeACPState` against a real `SessionDocument`, with `Date.now` stubbed
  rather than any clock being waited on: finalizing a turn, then finalizing the
  same session an hour later (what app close does), leaves the first `endedAt`
  in place; a turn that was never finished still gets stamped at teardown.
- New `apps/cli/src/lib/assistant-turn-finalize.test.ts` covers the rule itself
  (6 cases, two explicit `endedAt` values standing in for "turn ended" and "app
  closed"): open entry is stamped; a second stamp does not move the first; an
  entry with `finished: true` and no `endedAt` gains none; an explicit `turnId`
  is honoured past a trailing image entry; `permissionWaitMs` is recorded; a
  trailing non-assistant entry is never stamped.
- Both suites fail without the guard (3 failures) and pass with it (8 passed).
- `corepack pnpm format` — clean; all three new files pass `prettier --check`.
- `corepack pnpm check` does not complete on this machine: it aborts in its
  first step (`typecheck` -> `lody prepare:acp-adapters`) on a pre-existing
  `acp-extension-dsh` build error, `src/adapter.ts(1649,7) TS2352` on
  `ReadableStream` generics. That submodule is untouched here
  (`git diff --submodule` and `git status` are clean for it) and the local
  toolchain is Node v24.16.0, so this may be a local `@types/node` lib
  mismatch rather than something CI sees.
- Every other `check` stage was run individually: recursive `typecheck` with
  `acp-extension-dsh` excluded (covers `lody`, all packages, and `site-docs`)
  — pass; `corepack pnpm lint` (oxlint type-aware) — 0 errors;
  `corepack pnpm lint:i18n`, `check:code-collab-imports`,
  `check:platform-boundaries`, `check:public-boundary` — all pass.
- Rebased onto current `main` and re-verified there: applies with no conflict,
  `apps/cli` typechecks clean, both new suites pass (8 cases).
- `corepack pnpm test:ci` on the pre-rebase base: `apps/cli` 248 files / 2474
  tests and every package pass. That count predates both new suites; a full
  `apps/cli` run on the rebased branch reports 250 files / 2489 tests, the delta
  being the two new suites plus suites `main` gained during the rebase window.
  One `apps/electron` file, `src/main/services/loro-data-plane-relay.test.mjs`,
  fails on `Error: Electron failed to install correctly` — the Electron binary
  is missing from this machine's install, not a code failure, and nothing here
  touches `apps/electron`.
- That `test:ci` run predates both new suites, so they were also run together
  with the neighbouring `message-handler-*` suites (5 files, 29 tests, all pass).
  `apps/cli` `typecheck` was re-run afterwards; note its tsconfig excludes test
  files, so that step covers `assistant-turn-finalize.ts` but not either new
  test file — those are typechecked only by `vitest`.
- Not done: no end-to-end app-close run. The teardown path itself
  (`exit`/`terminated` -> `finalizeACPState`) is unchanged; only the history
  write it performs is guarded.

- **Review focus:** `apps/cli/src/lib/assistant-turn-finalize.ts` (one guard, extracted verbatim otherwise) and the `finalizeACPState` call site in `message-handler.ts`.
- **Decisions to challenge:** guarding on `finished === true` instead of a missing `endedAt`, and extracting the loop into its own module rather than leaving it inline.
- **Plausible failures / evidence gaps:** An adjacent bug is untouched: image/file entries are pushed after the turn entry, so a no-turnId finalize on a session ending with one matches that entry and breaks, and the real turn entry is never finalized at all — previously it got the wrong entry stamped, now neither is. Fixing it means scanning past finished entries, which would stamp `now` on genuinely abandoned older turns; that trade-off is yours, not this PR's. Also intended but worth confirming: when a teardown finalize precedes a turnId finalize, the first stamp wins and the later `permissionWaitMs` is dropped.

### Authoring context

- **User goal / directives:** Fix issue #260 ("Worked for" duration grows while the app stays open) at its root cause, with a regression test and no unrelated changes.
- **Constraints / non-goals:** No behaviour change for interrupted turns or resume; no touching the teardown call sites; no fix for the adjacent image/file-entry ordering bug; tests use no real sleeps or wall-clock races per `AGENTS.md`.
- **Risk-bearing decisions:** The guard makes the first terminal stamp final for an assistant entry, so any future producer that pre-marks a live turn `finished: true` would suppress its real timing; all current `finished: true` writers in `apps/cli/src` were checked and write `system`/`user` entries or separate assistant image/file entries.
- **Destructive or irreversible behavior:** None. No migration, deletion, or rewrite of stored history; the change only declines a write that previously overwrote a durable field.
- **Deliberately not done or tested:** The adjacent finalize-target bug described above, and a manual app-close reproduction — `finalizeACPState` is driven directly against a real `SessionDocument` instead, which is the boundary the bug lives at.
- **Unknowns / confidence:** High confidence in the root cause and the guard; the residual unknown is how often sessions end with an image/file entry, which decides how much of #260 the untouched adjacent bug still accounts for.

### DIFF
--- a/apps/cli/src/lib/assistant-turn-finalize.test.ts
+++ b/apps/cli/src/lib/assistant-turn-finalize.test.ts
@@ -1,0 +1,87 @@
+import { describe, expect, it } from 'vitest';
+import type { SessionHistoryInput } from '@lody/shared';
+
+import { markAssistantTurnFinished } from './assistant-turn-finalize';
+
+const OPENED_AT = Date.parse('2026-01-01T00:00:00.000Z');
+const TURN_ENDED_AT = OPENED_AT + 12_000;
+const APP_CLOSED_AT = OPENED_AT + 3_600_000;
+
+const assistantEntry = (
+  overrides: Partial<SessionHistoryInput> & { id: string }
+): SessionHistoryInput => ({
+  role: 'assistant',
+  timestamp: new Date(OPENED_AT).toISOString(),
+  items: [],
+  fileDiff: [],
+  ...overrides,
+});
+
+describe('markAssistantTurnFinished', () => {
+  it('stamps the open assistant entry when no turn id is given', () => {
+    const history = [assistantEntry({ id: 'assistant:u1' })];
+
+    markAssistantTurnFinished(history, { endedAt: TURN_ENDED_AT });
+
+    expect(history[0]).toMatchObject({ finished: true, endedAt: TURN_ENDED_AT });
+  });
+
+  it('leaves an already finished turn alone when a later teardown finalizes', () => {
+    // Regression for #260: closing the app runs the no-turnId finalize for every
+    // live session, which used to re-stamp `endedAt = now` on a turn that ended
+    // an hour earlier and inflate its rendered "Worked for …".
+    const history = [assistantEntry({ id: 'assistant:u1' })];
+
+    markAssistantTurnFinished(history, { endedAt: TURN_ENDED_AT });
+    markAssistantTurnFinished(history, { endedAt: APP_CLOSED_AT });
+
+    expect(history[0]).toMatchObject({ finished: true, endedAt: TURN_ENDED_AT });
+  });
+
+  it('records no duration for a finished entry that never carried one', () => {
+    // Image-group and file entries publish `finished: true` with no `endedAt`.
+    const history = [assistantEntry({ id: 'assistant-image-1', finished: true })];
+
+    markAssistantTurnFinished(history, { endedAt: APP_CLOSED_AT });
+
+    expect(history[0]?.endedAt).toBeUndefined();
+  });
+
+  it('stamps the addressed turn even when a later assistant entry follows it', () => {
+    const history = [
+      assistantEntry({ id: 'assistant:u1' }),
+      assistantEntry({ id: 'assistant-image-1', finished: true }),
+    ];
+
+    markAssistantTurnFinished(history, { turnId: 'assistant:u1', endedAt: TURN_ENDED_AT });
+
+    expect(history[0]).toMatchObject({ finished: true, endedAt: TURN_ENDED_AT });
+    expect(history[1]?.endedAt).toBeUndefined();
+  });
+
+  it('records the permission wait when the finalizing turn measured one', () => {
+    const history = [assistantEntry({ id: 'assistant:u1' })];
+
+    markAssistantTurnFinished(history, { endedAt: TURN_ENDED_AT, permissionWaitMs: 4_000 });
+
+    expect(history[0]?.permissionWaitMs).toBe(4_000);
+  });
+
+  it('never stamps a user or system entry standing after the turn', () => {
+    const history: SessionHistoryInput[] = [
+      assistantEntry({ id: 'assistant:u1' }),
+      {
+        id: 'system:1',
+        role: 'system',
+        timestamp: new Date(OPENED_AT).toISOString(),
+        items: [],
+        fileDiff: [],
+      },
+    ];
+
+    markAssistantTurnFinished(history, { endedAt: TURN_ENDED_AT });
+
+    expect(history[0]).toMatchObject({ finished: true, endedAt: TURN_ENDED_AT });
+    expect(history[1]?.finished).toBeUndefined();
+  });
+});
--- a/apps/cli/src/lib/assistant-turn-finalize.ts
+++ b/apps/cli/src/lib/assistant-turn-finalize.ts
@@ -1,0 +1,51 @@
+import type { SessionHistoryInput } from '@lody/shared';
+
+/**
+ * Stamp the terminal footprint (`finished`/`endedAt`/`permissionWaitMs`) on the
+ * assistant entry a finalize call owns. Extracted from `finalizeACPState` so the
+ * one rule that matters here is testable: a terminal stamp is written once.
+ *
+ * `finalizeACPState` has a no-turnId overload used by teardown/cancel paths
+ * (session `exit`/`terminated`, error, cleanup). Those callers only check that
+ * transient state exists, so on app close they run for sessions whose turn ended
+ * long ago — and the loop matched the last assistant entry regardless of state,
+ * re-stamping `endedAt = now`. The renderer derives "Worked for …" from
+ * `endedAt - timestamp`, so every close inflated a finished turn's duration by
+ * the wall-clock time the app stayed open.
+ *
+ * Skipping an already-finished entry (rather than filling in a missing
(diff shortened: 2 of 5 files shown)


Leeeon233:
LGTM

PR #410, opened 2026-09-05
fix(components): keep New Chat attachments across tab switches

### LINKED ISSUE #242: [Bug] New chat image attachment disappears after switching tabs
### Affected area

Desktop app

### Installation method

Desktop release

### Lody version or commit

0.89.0

### Operating system

macOS 27.0 arm64

### Agent or runtime

Not involved in this reproduction.

### What happened?

On the New Chat page, when the composer contains both an image attachment and text, switching to another tab or page in Lody and then returning leaves the text intact but removes the image attachment. No visible error is shown.

This can cause the intended image context to be silently omitted if the user sends the message without noticing that the attachment disappeared.

### What did you expect?

The image attachment and text should both remain in the New Chat draft until the user removes them, clears the draft, or sends the message successfully.

### How can we reproduce it?

1. Open the Lody desktop app and go to New Chat.
2. Add an image attachment to the composer.
3. Enter some text without sending the message.
4. Switch to another tab or page in Lody, such as an existing session.
5. Return to New Chat.
6. Observe that the text is still present but the image attachment is gone.

### How often does it happen?

Every time

### Relevant log output

No visible error or relevant log output is shown when the attachment disappears.

### Additional context

This is not limited to images. All attachment types disappear from the New Chat draft after switching away and returning, while the typed text remains.

Code inspection shows that the text and attachments have different state lifetimes:

- The text draft is stored by `chatLandingSessionStateAtomFamily` with `atomWithStorage`, so it is restored from `localStorage` after returning to New Chat.
- Image and file drafts are held in component-local state inside `useChatLandingImageDraft` and `useChatLandingFileDraft`. On unmount, the image cleanup revokes preview object URLs and clears image state, while the file cleanup aborts active transfers and clears file state.
- The reserved draft session ID is also component-local in `useChatLandingDraftSession`, so it is lost when `ChatLanding` unmounts.

On desktop, navigating away from `/chat` unmounts `ChatLanding`. When it mounts again, the text draft is rehydrated, but the attachment drafts and reserved session ID start empty. This state-lifetime mismatch explains the observed behavior.

### Before submitting

audichuang:
## Related issue

Closes #242

## Problem / pressure

On the New Chat landing, an image attached to the composer disappears when the
user visits another tab and comes back. The text stays. Nothing tells them the
picture is gone, so the message ships without the context it was written around.

The draft was split across two lifetimes. Prompt text lives in
`chatLandingSessionStateAtomFamily`, a module-level atom, so it survives the
chat route unmounting. Attachments and the reserved draft session id were
component `useState` / `useRef`, and the landing unmounts on every navigation
away — so they were gone. Worse, both attachment hooks ran an unmount cleanup
that made the loss irreversible even if the state had been kept: the image hook
revoked every preview `blob:` URL, and the file hook aborted every upload still
in flight.

## Summary

Both pending-attachment lists and the reserved draft session id move to
module-level atoms in a new `atoms/chat-landing-draft.ts`, keyed by
`buildChatLandingDraftKey`, so the draft outlives the route exactly as the text
does. Deliberately not `atomWithStorage`: a `blob:` URL and an `AbortController`
do not serialize, and losing a draft attachment when the app restarts is
expected.

With the state kept, the unmount cleanups had to go. `URL.revokeObjectURL` and
`AbortController.abort()` now belong only to the user's own actions — removing
one attachment, or clearing the whole draft on send-accepted / draft reset. An
upload still in flight when the user leaves keeps running and settles into the
atom, so returning shows the finished attachment. For images this costs nothing
new: `uploadSessionImage` takes no abort signal, so those requests already
outlived the unmount and only their result was being discarded.

The attachment key is workspace-scoped where the prompt-text key (`userId`) is
not, because an `imageId` / `fileId` is addressable only inside the workspace it
was uploaded to; carrying one into a session created elsewhere would attach a
block pointing at another workspace's object. It scopes on the workspace **slug**
rather than the resolved id: `useResolvedWorkspaceScope()` reports `null` until
the workspace resolves, and a key that flipped mid-mount would strand whatever
was added first.

`useChatLandingDraftSession` now reads and writes the store directly, which also
removes the state/ref pair its synchronous `ensureSessionId` needed.

One more thing the hoist forced: `chat-landing.tsx` guarded the `resetDraftKey`
clear with a `useRef`, which is per-mount. That was harmless while the state it
cleared was per-mount too, but a `New chat` URL keeps its `resetDraftKey` in the
history entry, so navigating back to it would re-apply the same reset and destroy
the very draft this change preserves — revoking its preview URLs and aborting its
uploads. The marker moves into `chatLandingAppliedResetKeyAtomFamily`, scoped
exactly like the draft it guards.

This is 515 additions + 84 deletions, over the 200-line size threshold. It does
not split usefully: the three hooks share one reserved session id, and shipping
the image half without the file half (or the state hoist without the cleanup
move) leaves a half-migrated draft that is worse than the bug. 320 of those
lines are the new test file; the production change is ~120 lines across six
files. The Issue is linked, as the policy requires.

## Before / after

| Before | After |
| ------ | ----- |
| Attach an image on New Chat, visit another tab, return — text remains, attachment is gone | Both remain until the user removes them, clears the draft, or sends |
| Leaving the landing revoked every preview `blob:` URL | Preview URLs are revoked when that image is removed, or the draft is cleared |
| Leaving the landing aborted every in-flight file upload | An in-flight upload keeps running and settles into the restored draft |
| The reserved draft session id was regenerated on every remount | The reserved id is stable across the round trip, so attachments and `startSession` stay on one identity |
| Attachment state was implicitly per-mount, so workspace scope never came up | Attachments are keyed per workspace; prompt text still follows the user across workspaces |

## Test plan

New `packages/components/tests/chat-landing-draft-persistence.test.tsx` mounts
the three hooks under a real `jotai` store and a real React root, then unmounts
and remounts to reproduce the tab switch. Six cases: images and the reserved
session id survive with the same preview URL; a preview URL is revoked when that
image is removed; an image upload in flight at unmount finishes into the restored
draft; a file upload's `AbortSignal` is not aborted by the unmount and its result
lands on remount; clearing the draft (what submit-accepted and `resetDraftKey`
call) still revokes the preview URL, aborts the upload, and stays cleared across
a remount; drafts in two workspaces stay separate. Uploads are injected deferred
promises resolved by explicit signals, and `URL.createObjectURL` is stubbed — no
timers, sleeps, or microtask counting. Re-running the suite with the two unmount
cleanups restored fails four cases, so the tests bind the behavior rather than
the implementation.

- Rebased onto current `main` and re-verified there: applies with no conflict,
  `@lody/components` typechecks clean, the new suite passes (6 cases).
- `pnpm --filter @lody/components test` — 420 files, 3019 tests. Four failures, none
  in a file this diff touches: `tests/markdown-streaming-reparse.test.ts` fails the same
  way on an unmodified `main` checkout, and `tests/path-launchers-setting.test.tsx` plus
  two `tests/agent-config-dialog.test.tsx` cases are 5s-timeout misses under a saturated
  box that pass in isolation (33/33). Reported rather than worked around.
- `pnpm lint` (oxlint, type-aware) — 0 errors.
- `pnpm lint:i18n`, `check:code-collab-imports`, `check:platform-boundaries`,
  `check:public-boundary` — all pass.
- Typecheck passes for every workspace package including `@lody/components` and
  `apps/cli`.
- `pnpm check` does not complete on this machine, for three reasons in packages
  this diff does not touch — every changed file is under `packages/components`,
  and each failing package's working tree is identical to `main`, so these were
  run against unmodified base content. The `packages/acp-extension-dsh`
  submodule fails `tsc` at `src/adapter.ts:1273` with `TS2352` on a
  `ReadableStream` cast; one `@lody/turn-diff-store` SQLite GC test exceeds its
  5s timeout; one `@lody/electron` test file needs an Electron binary that did
  not install in this sandbox. They are reported here rather than claimed as
  passing.
- Not exercised: a manual desktop run of the reproduction steps.

- **Review focus:** `use-chat-landing-image-draft.ts` and
  `use-chat-landing-file-draft.ts`, where the unmount cleanup effects were
  deleted — confirm every remaining `revokeObjectURL` / `abort()` path still
  fires on remove and on `clearPending*`, which submit and `resetDraftKey` call.
- **Decisions to challenge:** keying attachments on the workspace slug while the
  prompt text stays keyed on `userId` alone, which is a visible asymmetry across
  a workspace switch; and letting a file upload continue after the user leaves
  New Chat instead of aborting it.
- **Plausible failures / evidence gaps:** a draft abandoned in a visited
  workspace holds its `File` objects and unrevoked preview URLs for the life of
  the page (bounded at 8 images per workspace key), and an upload that finishes
  for a draft the user never sends leaves an unreferenced object in storage —
  both accepted for a draft that is meant to still be there. The `resetDraftKey`
  back-nav path has no test — the guard is in the component, the suite is
  hook-level — and is instead correct by construction, the marker now sharing the
  draft's scope. Verified by the new suite, not a manual desktop run.

### Authoring context

- **User goal / directives:** fix #242 so a New Chat attachment survives a tab
  switch the way the prompt text already does, and land it as one focused,
  tested change against `main`.
- **Constraints / non-goals:** no localStorage persistence for attachments; no
  change to the upload protocol, the composer UI, or the submit path; reuse the
  existing landing `stateKey` convention rather than inventing a second keying
  scheme; do not touch the in-session composer.
- **Risk-bearing decisions:** attachment state is now process-global rather than
  per-mount, so the key is what keeps two workspaces apart — it is workspace-
  scoped on purpose, and on the slug rather than the resolved id so it cannot
  flip mid-mount. Dropping the unmount abort means a file upload survives the
  user leaving the landing.
- **Destructive or irreversible behavior:** none added. The change removes two
  destructive unmount paths (revoking preview URLs, aborting uploads) and leaves
  those operations on the user-initiated remove and clear paths, which are
  unchanged. No migration, no persisted format, nothing to roll back beyond
  reverting the commit.
- **Deliberately not done or tested:** no auto-retry or resumable-upload state
  for an attachment that fails while the landing is unmounted — the existing
  retry button covers it. No `atomFamily` eviction: entries are one per
  user+workspace visited in a session. No manual desktop reproduction; the new
  suite covers the unmount/remount round trip instead.
- **Unknowns / confidence:** high on the fix and its scope — the failing case is
  reproduced and pinned by tests, the production change is ~90 lines, and the
  landing has exactly two mount sites in this repository (the desktop chat route
  and the mobile workspace stack, the latter unaffected because it keeps the
  landing mounted). The residual judgment call is the workspace-key asymmetry
  described above.

### DIFF
--- a/packages/components/src/atoms/chat-landing-draft.ts
+++ b/packages/components/src/atoms/chat-landing-draft.ts
@@ -1,0 +1,80 @@
+import { atom } from 'jotai';
+import { atomFamily } from 'jotai/utils';
+import type { SessionFilePayload, SessionId, SessionImagePayload } from '@lody/shared';
+import type { SessionFileTransferPhase } from '@/lib/session-file-upload';
+
+/**
+ * In-memory chat-landing draft state that must outlive the landing route's
+ * unmount. The prompt text already survives through
+ * `chatLandingSessionStateAtomFamily`; attachments and the reserved draft
+ * session id used to live in component state, so navigating to another tab and
+ * back silently dropped them (#242).
+ *
+ * Deliberately NOT `atomWithStorage`: a preview `blob:` URL and an in-flight
+ * upload's `AbortController` cannot be serialized. Surviving a route unmount is
+ * the whole requirement — losing a draft attachment when the app restarts is
+ * expected.
+ */
+
+export type PendingImage = {
+  localId: string;
+  previewUrl: string;
+  file: File;
+  status: 'uploading' | 'uploaded' | 'failed';
+  progress: number;
+  error?: string;
+  uploaded?: SessionImagePayload;
+};
+
+export type PendingFile = {
+  localId: string;
+  file: File;
+  status: SessionFileTransferPhase | 'uploaded' | 'failed';
+  progress: number;
+  error?: string;
+  uploaded?: SessionFilePayload;
+  abort?: AbortController;
+};
+
+/**
+ * The one home for the landing draft's attachment scope. Attachments are
+ * uploaded into a specific workspace — an `imageId`/`fileId` from one workspace
+ * cannot be attached to a session in another — so unlike the prompt text (keyed
+ * by user alone) the attachment draft is workspace-scoped.
+ *
+ * Keyed on the workspace SLUG rather than the resolved id: the slug is a route
+ * param that is stable for the whole mount, while `useResolvedWorkspaceScope()`
+ * reports `null` until the workspace resolves. A key that flips mid-mount would
+ * strand whatever was added before it settled.
+ *
+ * `stateKey` is the prompt text's own key — the one passed to
+ * `chatLandingSessionStateAtomFamily`, which an alternate landing surface may
+ * suffix so it does not clobber the main draft. Pass that, not a raw user id, or
+ * a suffixed surface would share this scope with the main landing.
+ */
+export const buildChatLandingDraftKey = (stateKey: string | null, workspaceSlug: string): string =>
+  `${stateKey ?? 'anonymous'}:${workspaceSlug}`;
+
+export const chatLandingPendingImagesAtomFamily = atomFamily((_draftKey: string) =>
+  atom<PendingImage[]>([])
+);
+
+export const chatLandingPendingFilesAtomFamily = atomFamily((_draftKey: string) =>
+  atom<PendingFile[]>([])
+);
+
+export const chatLandingDraftSessionIdAtomFamily = atomFamily((_draftKey: string) =>
+  atom<SessionId | null>(null)
+);
+
+/**
+ * The `resetDraftKey` this scope has already been cleared for. It lives beside
+ * the draft rather than in a mount-scoped ref because the draft now outlives the
+ * route: a `New chat` URL keeps its `resetDraftKey` in the history entry, so
+ * navigating back to it would otherwise re-apply the same reset and destroy the
+ * draft this module exists to preserve — revoking its preview URLs and aborting
+ * its uploads on the way out.
+ */
+export const chatLandingAppliedResetKeyAtomFamily = atomFamily((_draftKey: string) =>
+  atom<string | null>(null)
+);
--- a/packages/components/src/components/chat/AGENTS.md
+++ b/packages/components/src/components/chat/AGENTS.md
@@ -93,6 +93,22 @@
   that same identity. Attachment hooks never reset it independently; reset only
   after full draft clear. Submit blocks while either `hasBlockingImages` or
   `hasBlockingFiles`.
+- The reserved session id and both attachment lists live in module-level atoms
+  (`atoms/chat-landing-draft.ts`), keyed by `buildChatLandingDraftKey`, so the
+  draft survives the landing route unmounting when the user visits another tab
+  — the prompt text always did, and an attachment that silently vanished was
+  sent-without-context waiting to happen (#242). Consequences the hooks must
+  keep: NO unmount cleanup (revoking a preview URL or aborting an upload on the
+  way out is what broke the returning draft), so `URL.revokeObjectURL` and
+  `AbortController.abort()` belong only to removing one attachment or clearing
+  the whole draft; an upload in flight at unmount keeps running and settles into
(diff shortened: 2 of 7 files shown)


Leeeon233:
Nice!

PR #411, opened 2026-09-05
fix(cli): allow local session creation on stale Flock sync

### LINKED ISSUE #398: [Bug] Session create hard-fails on a best-effort Flock freshness sync when the network is degraded
### Affected area

CLI / daemon

### Installation method

Built from source

### Lody version or commit

lody-oss @ 48249c9, CLI 0.89.3

### Operating system

macOS (arm64)

### Agent or runtime

N/A — fails before any agent is started

### What happened?

While this machine's outbound network was degraded, **every** attempt to create a session through MCP `lody_session_create` was refused — even though the machine was online, the target local project was already known locally, and `lody_session_create_options` had just returned that same project successfully from local state moments earlier.

The failure comes from a *freshness confirmation* sync, not from anything the create actually needs:

```
COMMAND_REJECTED: Flock document sync failed for <ws>:mf:<machine>
  (session.local-projects:<machine>): Streams sync failed: network_error
```

Path:

```
lody_session_create
└─ resolveLocalProjectRefOrThrow            apps/cli/src/commands/session.ts:987
   └─ syncMachineFlockDocsForRead(…, 'session.local-projects')
                                            apps/cli/src/commands/session.ts:995
      └─ syncFlockDocOrThrow(…)             apps/cli/src/lib/loro/doc.ts:947
         └─ repo.sync({ scope: 'doc' })  →  Streams → network_error
```

`syncFlockDocOrThrow` throws after an 8s timeout (`LODY_LORO_SYNC_MACHINE_FLOCK_TIMEOUT_MS`), and the create is rejected before its Operation is ever persisted — a subsequent `lody_session_create({ operationId, resume: true })` returns `OPERATION_NOT_FOUND`.

Every input the create needs (`machineId`, `projectId`, prompt) is local. The remote round trip only confirms that the local Flock replica is fresh.

This also contradicts the documented intent one layer down. `apps/cli/AGENTS.md` says of dispatch:

> Cloud confirmation … is AWAITED … but **NEVER throws** — the durable pointer plus the SQLite Operation own delivery. **Do not reintroduce a hard-fail Streams ack on the dispatch write.**

The dispatch *write* is protected; this pre-dispatch *read* is not. The human CLI has an escape hatch for the equivalent read commands (`--offline`); the MCP surface has none.

### What did you expect?

With the machine online and the project locally known, session creation should proceed from the local replica and treat the Flock sync as best-effort (the same policy the dispatch write already uses), or expose an explicit offline equivalent on the MCP surface.

### How can we reproduce it?

1. Register a local project and connect the daemon to a workspace.
2. Degrade outbound network so Streams sync round trips exceed ~8s (packet loss or throttling is enough; a full disconnect also reproduces).
3. Call `lody_session_create_options` — it succeeds and returns the machine, agent config, and local project.
4. Call `lody_session_create` with exactly those ids.
5. It fails with `COMMAND_REJECTED: Flock document sync failed … Streams sync failed: network_error`.

### How often does it happen?

Every time

### Relevant log output

```text
2026-09-04T13:06:38.391Z [DEBUG] [W:<ws>] Machine Flock doc sync was not confirmed before continuing (machine=<machine> reason=<reason>): Streams sync failed: network_error
2026-09-04T13:07:22.162Z [DEBUG] [W:<ws>] Machine Flock doc sync was not confirmed before continuing (machine=<machine> reason=<reason>): Streams sync failed: network_error
```

For contrast, the same log window shows ordinary outbound request failures unrelated to Lody's own backend, confirming the trigger is local connectivity rather than a service outage:

```text
2026-09-04T13:09:23.632Z [DEBUG] [pr-poller] Network error polling <repo>; retrying after the attempt floor: fetch failed
```

### Additional context

The impact is disproportionate to the cause: a transient network problem makes session creation impossible, which is exactly the condition local-first execution is meant to survive.

Suggested fix, smallest first:

1. Resolve the local project from the local Flock replica and downgrade `syncMachineFlockDocsForRead` on the **create** path to best-effort (log, do not throw), matching `confirmDispatchSyncedBestEffort`.
2. If some caller genuinely needs the freshness guarantee, make it opt-in rather than the default, and give the MCP surface an explicit offline equivalent of the CLI's `--offline`.

Two related problems observed in the same window are filed separately: a self-sustaining Streams reconnect loop, and error-classification/retry-advice defects.

### Before submitting

UniversePeak:
## Related issue

Closes #398

## Problem / pressure

When the machine Flock freshness sync fails during degraded network conditions, MCP/local session creation rejects a project that is already available in the local replica. The remote round trip is a freshness check, not a prerequisite for resolving the local project needed to create the session.

## Summary

Keep the existing hard-fail sync helper for callers that require fresh reads, but use a best-effort wrapper for both local-project resolution paths. A failed freshness sync is logged and resolution continues from the local Flock replica.

## Visual explanation

Simple change: this PR only changes the failure policy at existing session-creation sync boundaries; the control flow remains the same, with failed freshness checks falling through to local replica resolution.

```text
resolveCreateContext
  ├─ resolveAgentConfigForCreate
  │    └─ best-effort session.agent-configs sync
  └─ resolveLocalProjectRefOnMachineOrThrow
       └─ best-effort session.local-projects sync
            └─ local replica resolution (preserve sync cause if unavailable)
```

## Before / after

| Before | After |
| ------ | ----- |
| A Flock freshness timeout/network error aborts local-project resolution and prevents session creation. | The failure is logged, local project resolution continues from the replica, and session creation can proceed with locally known inputs. |

## Test plan

- `corepack pnpm exec vitest run src/commands/session.test.ts --pool=forks --maxWorkers=1` — 65 tests passed.
- The regression forces `syncFlockDocOrThrow` to reject with `Streams sync failed: network_error` and verifies local project resolution still succeeds.
- `git diff --check` passed.
- Full integration verification against a live degraded Streams connection was not run.

### Original user prompt

Original user-reported request that authored this change (quoted verbatim from Issue #398, "[Bug] Session create hard-fails on a best-effort Flock freshness sync when the network is degraded"):

```text
While this machine's outbound network was degraded, **every** attempt to create a session through MCP `lody_session_create` was refused — even though the machine was online, the target local project was already known locally, and `lody_session_create_options` had just returned that same project successfully from local state moments earlier.

With the machine online and the project locally known, session creation should proceed from the local replica and treat the Flock sync as best-effort (the same policy the dispatch write already uses), or expose an explicit offline equivalent on the MCP surface.
```

- **Review focus:** Inspect `apps/cli/src/commands/session.ts` local-project resolution paths and confirm the best-effort wrapper does not alter agent-config or ACP capability freshness reads.
- **Decisions to challenge:** Confirm that CLI and MCP local-project resolution should share this policy because both consume the local replica before session creation.
- **Plausible failures / evidence gaps:** The regression covers the resolver boundary with a synthetic network error; a live degraded Streams integration test remains unavailable in this environment.

### Authoring context

- **User goal / directives:** Fix Issue #398 so locally resolvable session creation survives transient machine Flock freshness-sync failures.
- **Constraints / non-goals:** Keep the change focused; do not make unrelated Flock reads best-effort or add a new MCP offline option.
- **Risk-bearing decisions:** Only the two local-project resolution callers switch to best-effort; the original hard-fail helper remains unchanged for other freshness-sensitive reads.
- **Destructive or irreversible behavior:** None; the change only logs a failed freshness check and reads the existing local replica.
- **Deliberately not done or tested:** No live packet-loss/degraded-Streams integration test was run; the deterministic unit regression covers the failure boundary.
- **Unknowns / confidence:** High confidence in the local behavior and regression coverage; live transport behavior still belongs to CI/maintainer verification.

### DIFF
--- a/apps/cli/src/commands/session.test.ts
+++ b/apps/cli/src/commands/session.test.ts
@@ -1020,7 +1020,7 @@ describe('session command helpers', () => {
     ).toEqual([authorizedProject]);
   });
 
-  it('marks local project refs for worktree session creation', async () => {
+  it('continues local project resolution when machine Flock freshness sync fails', async () => {
     const rootPath = mkdtempSync(path.join(os.tmpdir(), 'lody-session-git-project-'));
     try {
       execFileSync('git', ['init'], { cwd: rootPath, stdio: 'ignore' });
@@ -1038,7 +1038,9 @@ describe('session command helpers', () => {
         ],
         { cwd: rootPath, stdio: 'ignore' }
       );
-      const syncFlockDocOrThrow = vi.fn(async () => undefined);
+      const syncFlockDocOrThrow = vi
+        .fn()
+        .mockRejectedValueOnce(new Error('Streams sync failed: network_error'));
       const manager = {
         syncFlockDocOrThrow,
         repo: {
--- a/apps/cli/src/commands/session.test.ts
+++ b/apps/cli/src/commands/session.test.ts
@@ -1020,7 +1020,7 @@ describe('session command helpers', () => {
     ).toEqual([authorizedProject]);
   });
 
-  it('continues local project resolution when machine Flock freshness sync fails', async () => {
+  it('marks local project refs for worktree session creation', async () => {
     const rootPath = mkdtempSync(path.join(os.tmpdir(), 'lody-session-git-project-'));
     try {
       execFileSync('git', ['init'], { cwd: rootPath, stdio: 'ignore' });
@@ -1038,9 +1038,7 @@ describe('session command helpers', () => {
         ],
         { cwd: rootPath, stdio: 'ignore' }
       );
-      const syncFlockDocOrThrow = vi
-        .fn()
-        .mockRejectedValueOnce(new Error('Streams sync failed: network_error'));
+      const syncFlockDocOrThrow = vi.fn(async () => undefined);
       const manager = {
         syncFlockDocOrThrow,
         repo: {
@@ -1099,6 +1097,69 @@ describe('session command helpers', () => {
     }
   });
 
+  it('continues local project resolution when machine Flock freshness sync fails', async () => {
+    const rootPath = mkdtempSync(path.join(os.tmpdir(), 'lody-session-git-project-'));
+    try {
+      execFileSync('git', ['init'], { cwd: rootPath, stdio: 'ignore' });
+      execFileSync(
+        'git',
+        [
+          '-c',
+          'user.name=Test',
+          '-c',
+          'user.email=test@example.com',
+          'commit',
+          '--allow-empty',
+          '-m',
+          'init',
+        ],
+        { cwd: rootPath, stdio: 'ignore' }
+      );
+      const syncFlockDocOrThrow = vi
+        .fn()
+        .mockRejectedValue(new Error('Streams sync failed: network_error'));
+      const manager = {
+        syncFlockDocOrThrow,
+        repo: {
+          getDocMeta: vi.fn(async () => ({
+            meta: {
+              localProjects: {
+                'local-project-1': {
+                  id: 'local-project-1',
+                  name: 'lody',
+                  rootPath,
+                  createdAtMs: 1,
+                },
+              },
+            },
+          })),
+          openFlockDoc: vi.fn(async () => ({
+            flock: {
+              scan: () => [],
+            },
+          })),
+        },
+      } as unknown as Parameters<typeof resolveLocalProjectRefOrThrow>[0];
+
+      await expect(
+        resolveLocalProjectRefOrThrow(
+          manager,
+          'workspace-1' as WorkspaceId,
+          'machine-id' as MachineId,
+          'lody'
+        )
+      ).resolves.toEqual({
+        kind: 'local',
+        localProjectId: 'local-project-1',
+      });
+      expect(syncFlockDocOrThrow).toHaveBeenCalledWith(
+        getMachineFlockDocId('workspace-1' as WorkspaceId, 'machine-id' as MachineId),
+        expect.objectContaining({ reason: 'session.local-projects:machine-id' })
+      );
+    } finally {
+      rmSync(rootPath, { recursive: true, force: true });
+    }
+  });
   it('does not synthesize a branch for non-git local projects', async () => {
     const rootPath = mkdtempSync(path.join(os.tmpdir(), 'lody-session-non-git-'));
     try {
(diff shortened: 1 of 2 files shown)


zxch3n (changes requested):
Thanks for taking #398 on — the direction matches the issue's first suggestion exactly, and the change itself is clean: no new imports, no new dependencies, typecheck passes, and the adjusted test passes locally (CI code checks did not run on this fork PR, so I ran them by hand).

Three things stand before this actually fixes #398, so I'm requesting changes rather than merging it as-is:

**1. The create path has a second hard-fail sync one step *earlier*, and this PR leaves it in place.**
`resolveCreateContext` calls `resolveAgentConfigForCreate` → `listAgentConfigsForMachine` → `syncMachineFlockDocsForRead(…, 'session.agent-configs')` (`apps/cli/src/commands/session.ts:2184`) *before* it reaches `resolveLocalProjectRefOnMachineOrThrow`. In the reported incident that earlier sync happened to succeed (the network was intermittent), which is why the failure surfaced at `session.local-projects`. Under sustained degradation the create still dies at `session.agent-configs`, so `Closes #398` is not yet true. The same best-effort treatment needs to cover that call.

**2. Remote-machine creates get a worse error, not a better one.**
`machineId` can name another machine. When this CLI has no local replica of that machine's Flock doc, the sync is the only source of its projects; after this change the failure becomes `Local project not found` instead of the sync error. Suggest one of: keep best-effort only when `machineId` is this machine, or when the project is not found after a failed sync, attach the sync failure to the not-found error so the cause is visible.

**3. Test coverage moved rather than grew.**
`marks local project refs for worktree session creation` was renamed and its mock changed to reject, so the success path (sync resolves, `useWorktree` marking) no longer has a test in that block. Please restore the original test and add the failure-path case as a new one.

If you'd rather land the current scope as a partial step, that's fine too — change `Closes #398` to `Refs #398` so the issue stays open for (1) and (2).

UniversePeak:
Addressed the requested changes in ba5253a6.

- Session creation now treats both agent-config and local-project Flock freshness syncs as best-effort, while preserving the hard-fail helper for ACP capability reads.
- If a failed sync leaves a project unavailable, the resulting error retains the sync cause for remote-machine diagnostics.
- Restored the worktree success regression and added a separate stale-sync success case.

Verification: `session.test.ts` — 66 tests passed; `git diff --check` passed. The PR remains open and is ready for re-review.

UniversePeak:
Added the required `## Visual explanation` section with a compact call-flow sketch covering both best-effort session-creation sync boundaries and preservation of the remote-machine sync cause. The policy checks are now green again; the code head remains `ba5253a`. This is ready for re-review.

PR #416, opened 2026-09-05
fix(electron): install Linux .deb updates without freezing the app

### LINKED ISSUE #415: [Bug] Linux .deb in-app update freezes the app while pkexec runs, and a failed install hides the retry
### Affected area

Desktop app

### Installation method

Desktop release

### Lody version or commit

0.91.0

### Operating system

Ubuntu 24.04.4 LTS x64, kernel 7.0.0-30-generic, GNOME 46 (Wayland), pkexec 124

### Agent or runtime

electron-updater 6.7.3; no agent or model is involved in this path.

### What happened?

Two separate failures on the Linux `.deb` update path. Both come from
`DebUpdater.doInstall` running its package manager through
`BaseUpdater.spawnSyncLog`, which uses `spawnSync`.

**1. The app freezes for as long as the privileged installer runs.**

`spawnSync` blocks the Electron main process, event loop included, until the
command exits. On Linux that command is `pkexec`, which does not exit until the
polkit password prompt is answered. Nothing on the JS side can interrupt a
`spawnSync` that has already started, so there is no timeout and no cancel.

Clicking **Update and restart** at 21:19:33 froze this machine's app. The
updater log ends at that click and never resumes, and the 30-minute update
checks stop with it. The prompt was on the GNOME session while the user was
attached over SSH, so nobody saw it for 51 minutes.

Separately, and not diagnosed: after the prompt was answered and `dpkg`
finished at 22:11:06, the `pkexec` parent still had not exited 70 minutes
later, with no child process left, so the main process stayed blocked. That may
be a polkit or pkexec-124 issue rather than anything Lody controls, but it is
what a blocking install exposes the app to:

```
$ ps -o pid,ppid,etime,cmd -p 2240080,2240081
    PID    PPID     ELAPSED CMD
2240080 1479882    02:01:13 /bin/sh -c pkexec --disable-internal-agent /bin/bash -c 'dpkg -i .../Lody-0.91.1-amd64.deb'
2240081 2240080    02:01:13 pkexec --disable-internal-agent /bin/bash -c dpkg -i .../Lody-0.91.1-amd64.deb
$ cat /proc/1479882/wchan     # Lody main, parent of the sh above
ep_poll
$ dpkg -l lody | tail -1
ii  lody           0.91.1       amd64
```

A window that stops repainting for the length of an unanswered prompt is
indistinguishable from a crash, and the app cannot report or cancel anything
while it is in that state.

**2. A failed install hides the only way to retry.**

`autoUpdater`'s `error` event sets `phase: 'error'`, which discards a package
that is still on disk and still installable:

- `readUpdateBannerState` returns `null` for any phase other than `downloading`
  or `downloaded`, so the sidebar update banner disappears.
- `about-setting.tsx` gates the install button on `isDownloaded`, so it is
  replaced by "Check for updates".
- `AppUpdaterService.quitAndInstall` rejects with `update_not_downloaded` for a
  package sitting in `~/.cache/@lodyelectron-updater/pending/`.

The recovery window is up to 30 minutes, until the next periodic check re-emits
`update-downloaded` from cache. The same hole swallows a downloaded update on
any transient network error during a periodic check.

This is not hypothetical: on 2026-09-04 `unattended-upgrades` held the dpkg
frontend lock, `dpkg -i` exited 2, the `apt-get install -f -y` fallback raised a
**second** password prompt and exited 100, and the update went unavailable. The
user recovered only by finding the manual **Check for updates** button first.

### What did you expect?

- The window keeps repainting and responding while the system password prompt
  is open, and the install can be cancelled.
- The app restarts only after the package manager actually succeeded.
- A failed install or a failed check leaves a downloaded update installable, so
  the button that retries it is still there.

### How can we reproduce it?

The freeze, on a Linux `.deb` install:

1. Install Lody from the `.deb` and let it download an update.
2. Click **Update and restart**.
3. Leave the polkit password prompt unanswered. (Equivalently: start the app
   where the prompt cannot be seen or answered, such as a session attached only
   over SSH.)
4. The window stops repainting and stops responding. `~/.config/Lody/updater.log`
   ends at `Executing: pkexec with args: ...` with no further lines, and the
   30-minute `Checking for update` entries stop.
5. Answer the prompt. The package installs and the app relaunches. (On this
   machine `pkexec` did not exit for another 70 minutes after `dpkg` finished,
   leaving the app blocked; that part is undiagnosed and may be polkit-side.)

The hidden retry, on the same install:

1. Hold the dpkg lock in another terminal (`sudo dpkg --configure -a`, or wait
   for an `unattended-upgrades` run).
2. Click **Update and restart**. `dpkg -i` exits 2, the `apt-get install -f -y`
   fallback raises a second password prompt and fails too.
3. The sidebar update banner disappears and the About panel shows
   **Check for updates** instead of **Update and restart**, for a package that
   is still in the pending cache. `updater.quitAndInstall` answers
   `update_not_downloaded`.
4. Press **Check for updates** manually, and the same install succeeds.

### How often does it happen?

Every time

### Relevant log output

```shell
# ~/.config/Lody/updater.log timestamps are UTC; system logs below are CST (UTC+8).
# 1. The freeze. Last line written; nothing after it, including the 30-minute checks.
2026-09-05T13:19:33.232Z [info] Install on explicit quitAndInstall
2026-09-05T13:19:33.233Z [info] Install: isSilent: false, isForceRunAfter: true
2026-09-05T13:19:33.268Z [info] Running as non-root user, using sudo to install: pkexec,--disable-internal-agent
2026-09-05T13:19:33.269Z [info] Executing: pkexec with args: --disable-internal-agent,/bin/bash,-c,'dpkg -i .../Lody-0.91.1-amd64.deb'

# The prompt was answered 51 minutes later. /var/log/dpkg.log (CST):
2026-09-05 22:10:31 upgrade lody:amd64 0.91.0 0.91.1
2026-09-05 22:11:06 status installed lody:amd64 0.91.1
# pkexec (no child left) still had not exited at 23:20 CST; undiagnosed, possibly polkit-side.

# 2. The hidden retry, one day earlier. dpkg lock held by another process:
2026-09-04T05:15:38.353Z [info] Executing: pkexec with args: --disable-internal-agent,/bin/bash,-c,'dpkg -i .../Lody-0.90.1-amd64.deb'
2026-09-04T05:15:42.641Z [error] dpkg: error: dpkg frontend lock was locked by another process with pid 1047887
2026-09-04T05:15:42.641Z [warn] Command pkexec exited with code 2
2026-09-04T05:15:42.641Z [warn] dpkg installation failed, trying to fix broken dependencies with apt-get
2026-09-04T05:15:42.654Z [info] Executing: pkexec with args: --disable-internal-agent,/bin/bash,-c,'apt-get install -f -y'
2026-09-04T05:15:46.316Z [error] E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 1047887 (dpkg)
2026-09-04T05:15:46.317Z [error] Error: Error: Command pkexec exited with code 100
    at DebUpdater2.spawnSyncLog (...)
    at DebUpdater2.runCommandWithSudoIfNeeded (...)
    at DebUpdater2.installWithCommandRunner (...)
    at DebUpdater2.doInstall (...)
    at AppUpdaterService.quitAndInstall (...)

# Manual re-check restored the button; the identical install then succeeded.
2026-09-04T05:15:52.405Z [info] Checking for update
2026-09-04T05:15:53.596Z [info] Install on explicit quitAndInstall
2026-09-04T05:15:53.617Z [info] Executing: pkexec with args: --disable-internal-agent,/bin/bash,-c,'dpkg -i .../Lody-0.90.1-amd64.deb'
```

### Additional context

**Ruled out.** The main process was checked directly: `NoNewPrivs: 0`. It was
started at 10:21:07 CST, 30 seconds after the previous update's `dpkg` finished
at 10:20:37, so it is the instance `app.relaunch()` produced, and its `pkexec`
did reach polkit. `/usr/bin/pkexec` is `-rwsr-xr-x root root`, the only copy on
disk. Neither failure involves privilege inheritance or a broken pkexec.

**Which platforms are affected.**

| install path | affected | why |
| --- | --- | --- |
| deb | yes | `DebUpdater.doInstall` -> `spawnSyncLog` -> `pkexec` |
| rpm / pacman | yes, identically (not currently shipped) | same shape, same helper |
| AppImage | no | rewrites its own file as the current user, no privileged helper |
| Windows / macOS | no | NSIS and Sparkle do not take this path |

**Relation to #277 / #278.** #278 made a failed install visible instead of a
permanent spinner, and it is why this was diagnosable at all: `updater.log`
exists because of it. It changed only the reporting path and said so. Both
failures above are in the install path it did not touch. The reporting fix also
depends on the freeze: it can read the recorded error synchronously after
`quitAndInstall` precisely because the whole app is blocked until the installer
exits.

**Suggested fix.** Upgrading does not help; electron-updater 6.8.9 still calls
`spawnSync` here, and `install()` is sync by design because it also runs from a
quit handler, which Lody does not use (`autoInstallOnAppQuit = false`). The
whole blocking path can be skipped for the one target that needs a privileged
helper:

1. On Linux, when the downloaded file is a `.deb` and the app is not an
   AppImage, spawn `pkexec --disable-internal-agent dpkg -i <file>`
   asynchronously and `app.relaunch()` / `app.quit()` only on exit code 0.
   Passing argv directly also removes the shell quoting the sync path needs,
   drops the `apt-get` retry that raises a second password prompt, and lets a
   signalled installer be reported as a failure rather than the success
   `spawnSync` reports for `status: null`.
2. Keep a downloaded package installable across a failed install or check, so
   the banner and the install button survive and the user can retry now instead
   of in 30 minutes.

A PR implementing both follows.

### Before submitting

audichuang:
## Related issue

Closes #415

## Problem / pressure

`DebUpdater.doInstall` runs its package manager through
`BaseUpdater.spawnSyncLog`, which uses `spawnSync`. That blocks the Electron
main process, event loop included, until the command exits, and on Linux the
command is `pkexec`, which does not exit until the polkit password prompt is
answered. Nothing on the JS side can interrupt a `spawnSync` already in
progress, so there is no timeout and no cancel.

On the reporter's machine one click froze the app for 51 minutes, waiting on a
prompt that was on a desktop session nobody was looking at. A window that stops
repainting that long is indistinguishable from a crash, and the app can neither
report nor cancel anything while it is in that state.

Separately, a failed install discarded a package that was still installable.
`phase: 'error'` makes `readUpdateBannerState` return `null` (sidebar banner
gone), makes `about-setting.tsx` swap the install button for "Check for
updates", and makes `quitAndInstall` answer `update_not_downloaded` for a file
sitting in the pending cache. A transient dpkg lock therefore cost the reporter
the retry until the next 30-minute check; they recovered only by finding the
manual re-check button.

## Summary

- New `app-updater-linux-install.ts` resolves the privileged command for a
  downloaded `.deb` and runs it asynchronously, resolving on the installer's
  exit rather than blocking the loop.
- `AppUpdaterService.quitAndInstall` is now `async`. When the plan resolves it
  spawns `pkexec --disable-internal-agent dpkg -i <file>` and calls
  `app.relaunch()` / `app.quit()` only once the installer has exited 0;
  otherwise it reports the exit code. Every other target keeps electron-updater
  unchanged.
- Argv is passed directly, so the shell quoting the sync path needed is gone,
  the `apt-get install -f -y` retry that raised a second password prompt for the
  same install is gone, and a signalled installer is a failure rather than the
  success `spawnSync` reports for `status: null`.
- `installInFlight` stops a second click from raising a second prompt now that
  the window stays interactive during the first.
- `recordError` keeps `phase: 'downloaded'` while a package is on disk, so a
  failed install or a failed periodic check no longer hides the retry. It
  replaces the two places that wrote `phase: 'error'` directly, and
  `quitAndInstall` detects an electron-updater install failure from an error
  count instead of from the phase.
- The About panel shows a destructive status line when a downloaded update
  carries an error, so the surface that keeps its button also says why the last
  attempt failed.

## Before / after

| Before | After |
| ------ | ----- |
| Main process blocked in `spawnSync` for the whole polkit prompt; app frozen, no cancel, no timeout | `spawn` + `await` on exit; the window keeps repainting and responding while the prompt is open |
| `app.relaunch()` ran whenever `doInstall` returned true, including for a signal-killed installer (`status: null`) | Relaunch only on exit code 0; signals, 126, 127 and any dpkg code are reported |
| `dpkg -i` failure retried through `apt-get install -f -y`, raising a second password prompt | Single privileged command; the dpkg error is reported as-is |
| A failed install or a network blip during a check set `phase: 'error'`, hiding the sidebar banner and the install button for up to 30 minutes | The package stays `downloaded` with its error attached; both surfaces keep the retry |
| About panel silently stopped its spinner on failure | About panel shows the failure next to the button it now keeps |

## Test plan

- `apps/electron/src/main/services/app-updater-linux-install.test.mjs` (new, 12
  cases, added to the package `test` script): plan resolution for deb vs
  AppImage vs non-Linux, a package path with spaces passed as one unescaped
  argument, and through an injected spawn — that the call is still unsettled
  after the spawn returns and settles only on `exit`, exit 0, 126, 127, a
  non-zero dpkg code with its stderr, a signalled installer, an `error` event,
  a spawn that throws, and `error` plus `exit` settling once.
- `pnpm --filter @lody/electron run test` — 90/90 pass excluding
  `loro-data-plane-relay.test.mjs`, which fails in this checkout with "Electron
  failed to install correctly": `node_modules/electron/dist/electron` is absent
  here, unrelated to this change.
- `pnpm --filter @lody/electron run typecheck`, `pnpm --filter @lody/components
  run typecheck`, `pnpm lint` (0 errors), `pnpm lint:i18n`,
  `pnpm check:public-boundary`, `pnpm check:platform-boundaries`,
  `pnpm check:code-collab-imports` — all pass.
- Full `pnpm typecheck` cannot complete in this checkout: `prepare:acp-adapters`
  fails building the `acp-extension-dsh` submodule on a `ReadableStream` type
  assertion. Verified identical on a clean tree with these changes stashed, so
  it is pre-existing and unrelated.
- Not verified: an end-to-end packaged `.deb` upgrade. It needs a signed release
  pair on the update feed, which this checkout cannot produce.

- **Review focus:** `app-updater-service.ts` `quitAndInstall` / `installLinuxDeb`
  for the quit-only-on-success ordering, and `recordError` plus the
  `update-available` / `update-downloaded` listeners for `downloadedFile`
  lifetime.
- **Decisions to challenge:** dropping the `apt-get install -f -y` fallback;
  reporting instead of falling back when `pkexec` is unavailable; keeping
  `phase: 'downloaded'` for every error while a package is on disk.
- **Plausible failures / evidence gaps:** no end-to-end packaged upgrade was
  run; the promise settles on the installer's `exit`, so a helper that never
  exits (observed once on the reporter's machine after a successful `dpkg`,
  undiagnosed) leaves the app responsive but not relaunched with a stuck
  spinner; `update-available` clears `downloadedFile` and `update-downloaded`
  re-sets it from cache each cycle, leaving a sub-second window where the plan
  does not resolve.

### Authoring context

- **User goal / directives:** Fix the Linux `.deb` in-app update at the root
  after a previous PR addressed only failure reporting, then file the issue and
  this PR.
- **Constraints / non-goals:** No dependency changes, no new IPC surface, no
  behavior change for Windows, macOS, AppImage, or Sparkle; the upstream
  `spawnSync` in electron-updater is bypassed, not patched.
- **Risk-bearing decisions:** Lody now owns quitting and relaunching for `.deb`
  installs, so a wrong exit-code reading would either skip a restart or restart
  without an install; both are reported to the renderer rather than silent.
- **Destructive or irreversible behavior:** `dpkg -i` on the downloaded package
  is unchanged from the previous path. Dropping the `apt-get -f` retry means an
  unmet-dependency install now stops and reports rather than attempting repair;
  recovery is `sudo apt-get install -f` and the error text says so.
- **Deliberately not done or tested:** rpm and pacman are not shipped and keep
  electron-updater; the mobile About surface is untouched because the bug is
  Linux-desktop only; no packaged end-to-end upgrade was run.
- **Unknowns / confidence:** High for the async install and the retained-package
  behavior, both covered by unit tests and read against electron-updater 6.7.3
  sources. Lower for exotic Linux sessions without a polkit agent, which now get
  a reported failure instead of a freeze but still cannot install in place.


Leeeon233:
Thank you so much!

PR #443, opened 2026-09-06
fix(shared): keep new history metadata primitive without migrating old data

zxch3n:
## Related issue

Same-repository follow-up for conversation storage size and compatibility.

## Problem / pressure

History catchall inference creates independent LoroText containers for ordinary metadata strings. Turning it off alone replaces legacy Text containers on edit, so the insertion policy needs a compatibility fix as well as explicit streaming fields.

## Summary

- Store newly inserted metadata (IDs, status, paths, commands and nested metadata) as primitive strings. Explicitly retain Text for message text/thought, plan markdown, tool text/output and script output.
- Pin a narrow Mirror 2.3.1 patch: string edits in inferred/Any or declared Text map fields retain the actual existing Text or primitive representation. Existing Text IDs survive edits and concurrent updates. No data migration runs.
- Add synthetic regression tests plus a read-only all-workspace storage audit using the real unpatched baseline Mirror engine and schema, not only old options on the new engine.
- Retain the candidate closed filtering schema and corpus calibration tool from this PR. Image/unknown-field filtering is still not wired into production writes.

## Before / after

| Case | Before | After |
| --- | --- | --- |
| New nested metadata string | Often a Text container | Primitive string |
| New streaming content | Inferred or declared Text | Explicit Text |
| Existing stored string | Schema change could replace its representation on edit | Existing representation and Text ID retained |
| Opening existing documents | Normal read | Same read; no migration or rewrite |

## Test plan

- Full shared suite: 1,045 tests across 92 files pass; six focused storage cases cover plain metadata, streaming fields, legacy snapshots, legacy primitives, update exchange and concurrent legacy Text edits.
- CLI ACP history: 22 tests pass against the updated shared package.
- Shared typecheck, focused Oxlint, formatting, public-boundary checks, patch applicability and lockfile hash checks pass. Full application build/typecheck and browser/mobile UI tests were not run locally.
- Final read-only corpus: all 7 workspace databases and 1,103 stored documents scanned; 323,883 message items reconstructed through the production new-write schema with exact JSON equality. Zero failures.
- Baseline/new reads match without adding operations; all reconstructed snapshots are readable by the unpatched old Mirror; old/new writer update exchange matches. 954 documents contain a legacy title Text tested for ID preservation on edit.
- SQLite is opened read-only with query_only enabled. All experimental writes are confined to disposable imported/new LoroDoc instances. No original database, snapshot, document or remote room is rewritten.
- Source documents' full root/turn metadata is compared on read. Reconstruction uses every message item in a minimal id/role/timestamp turn envelope; it is not a whole-document migration simulation.

- **Review focus:** Check the explicit streaming paths in schema.ts, both runtime/source halves of the Mirror patch, and the existing-Text concurrent-edit regression.
- **Decisions to challenge:** The boundary between primitive metadata and streaming Text, especially future provider fields, needs source-grounded review.
- **Plausible failures / evidence gaps:** The patch affects other inferred/Any and declared Text map string edits too; explicit transforms and non-Text declarations retain their previous behavior. Browser/mobile integration remains CI/manual scope.

### Authoring context

- **User goal / directives:** Reduce unnecessary Text containers only for new writes, preserve existing storage, and validate all locally stored documents plus backward compatibility.
- **Constraints / non-goals:** No migration, compaction, deletion, captured transcript fixtures, or production integration of candidate filtering.
- **Risk-bearing decisions:** New inserts use the new policy; existing string representations remain authoritative. Old unpatched writers can still create Text using their old policy, and new readers tolerate that.
- **Destructive or irreversible behavior:** None on existing user data. Application behavior changes only for new string insertion and compatible edits; audit mutations are disposable.
- **Deliberately not done or tested:** Image filtering rollout, attachment externalization, old-history compaction, future ConversationView integration and 3000-round UI performance remain separate work.
- **Unknowns / confidence:** Corpus and synthetic tests cover the pinned Loro 1.15.1/Mirror 2.3.1 baseline and the patched runtime, not every historical released application or future schema variant.

### DIFF
--- a/packages/components/src/components/AGENTS.md
+++ b/packages/components/src/components/AGENTS.md
@@ -1,7 +1,6 @@
 # Product surfaces (`src/components`)
 
-Parent `AGENTS.md` files also apply. `CLAUDE.md` is a symlink; edit `AGENTS.md` only.
-Child directories (`sessions/`, `mobile/`, `chat/`, `settings/`, …) own their own rules.
+Parent and child-directory `AGENTS.md` rules apply. Edit `AGENTS.md`, not `CLAUDE.md`.
 
 ## Sidebar and session rows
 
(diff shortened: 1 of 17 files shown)


zxch3n:
Superseded by #584, which reimplements the new-write storage policy on the current main architecture: primitive metadata, explicit streaming Text fields, and preservation of existing stored representations. Closing this older implementation to avoid duplicate integration. Payload externalization and on-demand retrieval remain separate unfinished work.

PR #446, opened 2026-09-06
fix(cli): persist task lifecycle progress into session history

### LINKED ISSUE #445: [Bug] Task lifecycle progress never reaches history, so the panel cannot show the running tool
### Affected area

Agent runtime / ACP

### Installation method

Built from source

### Lody version or commit

f90842d6

### Operating system

Linux 7.0.0-30-generic (Ubuntu)

### Agent or runtime

Claude Code — bundled `acp-extension-claude` 0.70.0 on `@anthropic-ai/claude-agent-sdk` 0.3.258

### What happened?

The subagent task panel's "Running {{tool}}" line can never appear, for any provider. Every non-terminal task lifecycle snapshot is dropped before it reaches persisted history, so the only field that carries the running tool name never arrives.

The chain, verified on `origin/main` at f90842d6:

1. `last_tool_name` reaches Lody on exactly one message. `grep -n "last_tool_name"` over the bundled `@anthropic-ai/claude-agent-sdk/sdk.d.ts` returns a single hit, line 5160, inside `SDKTaskProgressMessage`. `SDKTaskStartedMessage`, `SDKTaskUpdatedMessage` and `SDKTaskNotificationMessage` do not declare it.

2. The bundled adapter emits that as a non-terminal `tool_call_update`. `packages/acp-extension-claude/src/acp-agent.ts:242-245` maps `task_progress` to `status: "in_progress"`, and the object built at `:300-309` has `sessionUpdate: "tool_call_update"`, `_meta: { lody: { task } }`, and **no `rawInput` key at all**. `lastToolName` rides inside that `task` payload (`:274`, `:296`). The adapter's own test asserts this shape: `packages/acp-extension-claude/src/tests/acp-agent.test.ts:8727` ("emits task lifecycle notifications with full task payloads") feeds a `task_progress` carrying `last_tool_name: "Read"` at `:8749` and asserts the four emitted notifications are `["tool_call", "tool_call_update", "tool_call_update", "tool_call_update"]`.

3. `filterNotificationsForHistory` drops it. `apps/cli/src/lib/acp/history.ts:835-861` keeps a `tool_call_update` only when `status === 'completed' || status === 'failed'` (`:839`), or it carries a non-empty `rawInput` (`:844-847`), or it carries `_meta.claudeCode.toolResponse` (`:851-859`); otherwise `:861 return false`. `_meta.lody.task` is not a keep-reason, so the update never reaches `applyNotificationOnHistory`.

4. The panel reads only from persisted history. `packages/components/src/components/ai-gui/subagent-task-panel.tsx:88-91` renders `Running {{tool}}` from `task.lastToolName`, fed exclusively by `collectSubagentTasks(message.items)` (`view.tsx:807`, `:3381`), where `message` is the persisted Loro history. There is no client-side applier: `grep -rn "applyNotificationOnHistory\|parseLodyTaskMeta\|handleACPUpdateMessage" packages/components/src apps/electron/src` returns zero hits.

The same drop also loses a non-terminal `task_updated` snapshot, so mid-flight status changes and in-progress `usage` are lost too.

Why this survived: the only existing coverage, `apps/cli/src/lib/acp/history.test.ts:233` ("upserts Claude task lifecycle progress into a single subagent_task item"), exercises the legacy `rawInput.lodyClaudeTaskLifecycle` carrier, which passes the non-empty-`rawInput` branch. The shipped `_meta.lody.task` carrier has no history coverage at all.

### What did you expect?

A task lifecycle snapshot should reach history and update the existing `subagent_task` item, so the panel can show the running tool while a subagent works, instead of a static "Working…" until the task settles.

The filter's stated reason does not apply to this carrier. Its comment says persisting all intermediate snapshots "causes the CRDT history to blow up" — that concern is about per-tick payload bytes from terminal-output snapshots. A task snapshot is a handful of small scalars, and the applier upserts it in place by `taskId` (`packages/shared/src/acp/history-apply.ts:1737-1749`), so a kept tick is a few `LoroMap.set` operations on one item.

### How can we reproduce it?

This was reproduced at the code and unit level, not by observing the desktop UI — stated plainly so the report is not read as a UI observation.

Add a test to `apps/cli/src/lib/acp/history.test.ts` that drives `handleACPUpdateMessage` with the shape the bundled adapter emits for `task_progress`, then read the persisted history:

```
sessionUpdate: 'tool_call_update'
toolCallId:    'task:task-1'
kind:          'think'
status:        'in_progress'
_meta:         { lody: { task: { version: 1, taskId: 'task-1', kind: 'subagent',
                 status: 'in_progress', actor: 'Explore', lastToolName: 'Read',
                 usage: { totalTokens: 100, toolUses: 2, durationMs: 500 } } } }
```

No `subagent_task` item gains `lastToolName` or `usage`. Sending the identical `_meta.lody.task` payload on a `tool_call` (which is what `task_started` becomes) does persist, because the filter's `tool_call_update` branch is never reached for it — which isolates the filter as the only difference.

### How often does it happen?

Every time

### Additional context

Scope note: the sibling path `apps/cli/src/agent/claude-task-lifecycle.ts`, reached through the `_claude/taskLifecycle` extension method, is documented as "One-release compatibility for pre-Core-v0.1 managed runtimes" (`apps/cli/src/agent/lody-acp-extension.ts:181`). `grep -rn "taskLifecycle" packages/acp-extension-claude/src/` returns zero non-test hits, so the bundled adapter never uses that lane and `_meta.lody.task` is the live carrier.

### Before submitting

audichuang:
## Related issue

Closes #445

## Problem / pressure

The subagent task panel's "Running {{tool}}" line can never appear, for any provider.

`last_tool_name` reaches Lody on exactly one message — the SDK declares it on `SDKTaskProgressMessage` and nowhere else. The bundled adapter emits that as an `in_progress` `tool_call_update` whose only payload is `_meta.lody.task`, with no `rawInput` key at all (`packages/acp-extension-claude/src/acp-agent.ts:242-245`, `:300-309`; its own test pins the shape at `src/tests/acp-agent.test.ts:8727`). `filterNotificationsForHistory` keeps a `tool_call_update` only when terminal (`:839`), when it carries a non-empty `rawInput` (`:844-847`), or when it carries `_meta.claudeCode.toolResponse` (`:851-859`) — line numbers at the base commit f90842d6, since this section describes the pre-fix state — so every non-terminal task snapshot was dropped before `applyNotificationOnHistory`. The panel reads only persisted history (`view.tsx:807`, `:3381`; no client-side applier exists), so the field never arrived.

The same drop also lost non-terminal `task_updated` snapshots, so mid-flight status and in-progress `usage` never landed either.

This survived because the only existing coverage (`history.test.ts:233`) exercises the legacy `rawInput.lodyClaudeTaskLifecycle` carrier, which passes the non-empty-`rawInput` branch. The shipped `_meta.lody.task` carrier had no history coverage.

## Summary

One guard at the shared predicate, immediately before the final `return false`:

```
if (parseLodyTaskMeta(meta) !== null) return true;
```

Reusing the applier's own parser — the exact function `history-apply.ts:1106` calls — makes the predicate "kept iff it will materialize a `subagent_task` item". A malformed or future-shaped carrier is still dropped, and no update is kept that produces nothing.

Non-terminal snapshots therefore reach `upsertSubagentTask`, so a settled row now ignores them outright — one early return rather than pinning fields one by one. `task_progress` reports `in_progress` and re-derives `taskKind`/`isBackgrounded` from a registry the producer has already pruned, so a late tick would restart a finished spinner and badge a completed subagent Background.

`actor` needs a separate guard, and that hazard predates this change. It is derived from `subagent_type` / `workflow_name` / `task_type` — which the producer reliably sees only on `task_started`; every other event falls through to a `"Claude task"` placeholder. Being materialized unconditionally it is never absent, so the merge's absence-based preservation cannot protect it, and the terminal event already replaced a workflow's own name with the placeholder. Keeping the first value repairs that and stops non-terminal ticks doing the same mid-run. Every other field still lets the later event win; the pre-existing legacy-carrier test pins that a progress tick's `description` reaches the settled row.

The gate is recorded in `apps/cli/src/lib/acp/AGENTS.md`, since a new `_meta.lody.*` carrier needing mid-flight persistence has to know it must ride a `tool_call` or a terminal update.

The filter's blow-up warning does not transfer to this carrier, and the code comment says why rather than restating the code. The concern is per-tick payload **bytes** in the oplog, not item count — `upsertToolCall` merges in place too. What differs is the payload: a task snapshot is a handful of small scalars, so a kept tick is a few `LoroMap.set` ops on the single `subagent_task` item `upsertSubagentTask` keeps per `taskId`, not the repeated full-string replacement a terminal-output snapshot costs.

### Before / after

| Carrier | Before | After |
| --- | --- | --- |
| `task_started` → `tool_call` | persisted | persisted (unchanged) |
| `task_progress` → in-progress `tool_call_update` with `_meta.lody.task` | dropped | persisted, merged into the same item |
| non-terminal `task_updated` with `_meta.lody.task` | dropped | persisted, merged |
| terminal `task_notification` | persisted | persisted (unchanged) |
| in-progress `tool_call_update` with no task snapshot | dropped | dropped (unchanged) |
| a workflow's persisted `actor` once settled | `"Claude task"` (the terminal event clobbered it) | its own name |
| a progress tick arriving after the terminal event | dropped | ignored — no status, badge or tool-name churn on a settled row. **Defensive: no producer is known to emit this ordering.** It is guarded because the filter change is what would let one through |
| a mid-run row with no `lastToolName` | `Working…` | the task's `summary`, which every real `task_progress` carries and which now reaches history (`subagent-task-panel.tsx:93`) |
| any settled task's description column | the task's purpose, e.g. `verify task lifecycle reaches history` | the last progress tick's activity, e.g. `Check: checker`. `description` is required on every `task_progress`, so this is not workflow-only: a subagent row settles on its last activity string too. The terminal event carries no `description` to restore the purpose |

## Test plan

Ran on Linux, Node 22, at f90842d6 plus this commit.

- `pnpm --filter lody run test src/lib/acp/history.test.ts` — 29 passed (seven added; the file's other tests are untouched).
- Red-before/green-after demonstrated, not asserted: `git stash push -- apps/cli/src/lib/acp/history.ts` leaves only the tests in the tree; the suite then reports **4 failed / 25 passed**. Restoring them returns 29/29.
- One added test is a negative control that passes in **both** states ("keeps dropping an in-progress tool_call_update whose `_meta` carries no task snapshot"), so the guard is shown not to have widened past its intent.
- The identity regression is pinned against the SHIPPED lane: the five `_meta.lody.task` notifications of one workflow run, transcribed verbatim from a capture of the bundled adapter driven over ACP (Claude Code 2.1.258, the pinned runtime — its `system/init` confirms the version). Only the first carries an identity source; every later one says `"Claude task"`. Reverting the merge guard turns it red.
- `packages/shared/tests/acp-history-apply.test.ts` gains two tests so the applier change is covered in its own package; both fail against the base applier.
- `pnpm --filter @lody/shared run test` — 1030 passed.
- `pnpm --filter lody run test` — full package: **2503 passed, 1 skipped, 0 failed**.
- `pnpm --filter lody run typecheck` — clean.
- `pnpm lint` — 0 errors (9822 pre-existing warnings, unchanged by this diff).
- `pnpm check:public-boundary` — passed, 4207 files / 22 manifests.

Not run, and why: no desktop-UI verification. `loro-data-plane-relay.test.mjs` is the only electron-runtime test and cannot run in this checkout (the electron binary was never downloaded), and the recursive `pnpm check` chain stops earlier on an unrelated submodule build, so packages were checked individually. The defect and the fix were established at the code and unit level; the panel-rendering half is inferred from `subagent-task-panel.tsx:88-91` reading `task.lastToolName`, not observed in a running app.

- **Review focus:** `upsertSubagentTask` in `packages/shared/src/acp/history-apply.ts`. Two rules sit there: a settled row drops non-terminal snapshots wholesale, and `actor` is pinned on every event. Judge whether they belong at the call site or in the merge.
- **Decisions to challenge:** `description` was left later-wins. The argument for pinning `actor` applies to it verbatim; the only evidence the other way is `history.test.ts:233`, a maintainer-written test on a carrier with no producer in the tree. It was followed as the sole expressed intent rather than overridden — overrule it if that reading is wrong.
- **Plausible failures / evidence gaps:** the settled-row guard is defensive — no producer is known to emit a non-terminal snapshot after a terminal one. A completed subagent can still be badged Background by a route this PR does not touch: the adapter prunes its registry on the terminal `task_updated`, so the following `task_notification` re-derives `kind: 'background'`. No running-app check; the panel half is inferred from source.

### Authoring context

- **User goal / directives:** fix the task panel's missing running-tool line at its root cause, in the public tree, with a change the author would be willing to defend in review.
- **Constraints / non-goals:** no submodule edits (all `packages/acp-extension-*` are submodules); no change to `LodyTaskKind`, `LodyTaskStatus`, or any `acp-extension-core` contract; no widening of `LodyTaskMetaSchema` — an earlier attempt at that was dropped after review showed the lane it fixed has no shipped producer.
- **Risk-bearing decisions:** more notifications now persist into session history. Bounded by the in-place upsert at `history-apply.ts:1737-1749`, which keeps one `subagent_task` item per `taskId`; the per-tick cost is a few small `LoroMap.set` ops.
- **Destructive or irreversible behavior:** none — no migration, no rewrite of existing history, no deletion. Sessions recorded before this change simply keep the fields they already have.
- **Deliberately not done or tested:** the desktop UI was not launched; `packages/components` was not modified, so the panel is exercised only through the shape of the stored item. The sibling legacy lane (`apps/cli/src/agent/claude-task-lifecycle.ts`) was not edited, but its non-terminal ticks now persist too, through the same guards. A pre-existing clobber of `actor` at the terminal event is repaired by the same guard rather than left half-fixed.
- **Unknowns / confidence:** high confidence in the drop and the fix (each hop read on the shipped path, plus red-before/green-after). Lower confidence that the rendered line is the only user-visible improvement — mid-flight `usage` and status now persist too, and those surfaces were not inspected.

### DIFF
--- a/apps/cli/src/lib/acp/AGENTS.md
+++ b/apps/cli/src/lib/acp/AGENTS.md
@@ -74,3 +74,10 @@ first `await awaitTurnHistoryGate(sessionId)` — see
 `../../session/turn-history-gate.ts`. Add the same await to any NEW code path that
 appends or positions history entries during a turn; map-keyed status/meta writes stay
 ungated.
+
+A non-terminal `tool_call_update` reaches history only when it carries a non-empty
+`rawInput`, `_meta.claudeCode.toolResponse`, or a parseable `_meta.lody.task` —
+`filterNotificationsForHistory` in `history.ts`. A new `_meta.lody.*` carrier that needs
+mid-flight persistence must therefore ride a `tool_call` or a terminal update, or add its
+own clause there; otherwise it is dropped before the applier and unit tests that fabricate
+history will not catch it.
--- a/apps/cli/src/lib/acp/history.test.ts
+++ b/apps/cli/src/lib/acp/history.test.ts
@@ -372,6 +372,380 @@ describe('handleACPUpdateMessage', () => {
     });
   });
 
+  // The bundled Claude adapter publishes task lifecycle through `_meta.lody.task` on a
+  // `tool_call` (task_started) and non-terminal `tool_call_update`s (task_progress /
+  // task_updated). The fixtures below transcribe that update shape verbatim: no `rawInput`
+  // and no `_meta.claudeCode`, because that absence is what used to drop them from history.
+  const lodyTaskMeta = (task: Record<string, unknown>) => ({
+    lody: { task: { version: 1, kind: 'subagent', ...task } },
+  });
+
+  it('persists the shipped _meta.lody.task progress snapshot so lastToolName reaches the panel', async () => {
+    const { doc, readHistory } = createDoc();
+
+    await handleACPUpdateMessage(
+      doc,
+      [
+        parseSessionNotification({
+          sessionId: 'acp-session',
+          update: {
+            sessionUpdate: 'tool_call',
+            toolCallId: 'task:task-1',
+            title: 'Explore: Find CLI startup behavior',
+            kind: 'think',
+            status: 'in_progress',
+            _meta: lodyTaskMeta({
+              taskId: 'task-1',
+              status: 'in_progress',
+              actor: 'Explore',
+              description: 'Find CLI startup behavior',
+            }),
+          },
+        }),
+        parseSessionNotification({
+          sessionId: 'acp-session',
+          update: {
+            sessionUpdate: 'tool_call_update',
+            toolCallId: 'task:task-1',
+            title: 'Explore: Find CLI startup behavior',
+            kind: 'think',
+            status: 'in_progress',
+            _meta: lodyTaskMeta({
+              taskId: 'task-1',
+              status: 'in_progress',
+              actor: 'Explore',
+              description: 'Find CLI startup behavior',
+              lastToolName: 'Read',
+              usage: { totalTokens: 100, toolUses: 2, durationMs: 500 },
+            }),
+          },
+        }),
+      ],
+      { getCurrentSessionTurnId: () => 'turn-1' }
+    );
+
+    const items = readHistory()[0]?.items as MessageContent[] | undefined;
+    const task = items?.find(
+      (item): item is Extract<MessageContent, { type: 'subagent_task' }> =>
+        item.type === 'subagent_task'
+    );
+    expect(task).toMatchObject({
+      taskId: 'task-1',
+      status: 'in_progress',
+      actor: 'Explore',
+      description: 'Find CLI startup behavior',
+      lastToolName: 'Read',
+      usage: { totalTokens: 100, toolUses: 2, durationMs: 500 },
+    });
+    // Lifecycle events never leak a raw tool call into history.
+    expect(items?.some((item) => item.type === 'tool_call')).toBe(false);
+  });
+
+  it('merges repeated _meta.lody.task progress into a single subagent_task item', async () => {
+    const { doc, readHistory } = createDoc();
+    // `task_progress.description` is required and activity-scoped — it changes on
+    // essentially every tick — so the fixture changes it too.
+    const progress = (lastToolName: string) =>
+      parseSessionNotification({
+        sessionId: 'acp-session',
+        update: {
+          sessionUpdate: 'tool_call_update',
+          toolCallId: `task:task-2`,
+          title: `Explore: running ${lastToolName}`,
+          kind: 'think',
+          status: 'in_progress',
+          _meta: lodyTaskMeta({
+            taskId: 'task-2',
+            status: 'in_progress',
+            actor: 'Explore',
+            description: `running ${lastToolName}`,
+            lastToolName,
+          }),
+        },
+      });
+
+    await handleACPUpdateMessage(
+      doc,
+      [
+        parseSessionNotification({
+          sessionId: 'acp-session',
+          update: {
+            sessionUpdate: 'tool_call',
+            toolCallId: 'task:task-2',
+            title: 'Explore: Auditing history',
+            kind: 'think',
+            status: 'in_progress',
+            _meta: lodyTaskMeta({
+              taskId: 'task-2',
+              status: 'in_progress',
+              actor: 'Explore',
+              description: 'Auditing history',
+            }),
+          },
+        }),
+        ...['Read', 'Bash', 'Grep'].map(progress),
+        parseSessionNotification({
+          sessionId: 'acp-session',
+          update: {
+            sessionUpdate: 'tool_call_update',
+            toolCallId: 'task:task-2',
+            title: 'Explore: Auditing history',
+            kind: 'think',
+            status: 'completed',
+            _meta: lodyTaskMeta({
+              taskId: 'task-2',
+              status: 'completed',
+              actor: 'Explore',
+              summary: 'Audit done',
+            }),
+          },
+        }),
+      ],
+      { getCurrentSessionTurnId: () => 'turn-1' }
+    );
+
+    const items = readHistory()[0]?.items as MessageContent[] | undefined;
+    const tasks = items?.filter((item) => item.type === 'subagent_task') ?? [];
+    // Keeping every tick costs no extra item: the applier merges in place by taskId.
+    expect(tasks).toHaveLength(1);
+    expect(tasks[0]).toMatchObject({
+      taskId: 'task-2',
+      status: 'completed',
+      summary: 'Audit done',
+      // a progress tick's activity description wins and survives the terminal event,
+      // matching the legacy-carrier expectation pinned above
+      description: 'running Grep',
+      // the last progress tick wins, and survives the terminal event that omits it
+      lastToolName: 'Grep',
+    });
+  });
+
+  it('keeps a real workflow run named through the shipped adapter stream', async () => {
+    const { doc, readHistory } = createDoc();
+    // The six `_meta.lody.task` notifications of one workflow run, transcribed from a
+    // capture of the bundled adapter driven over ACP (Claude Code 2.1.258, the pinned
+    // runtime). Only `task_started` carries an identity source, so every later event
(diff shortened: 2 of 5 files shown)


audichuang:
Closing in favour of a single branch that carries this commit unchanged plus the follow-up work, so the review happens in one place.

Thank you for the three review findings — all of them reproduce, and here is where each stands:

- **Resumes swallowed by the settled-row guard** — fixed. `resumeAgent` / `sendInput` on a receiver that is `running` or `pendingInit` now publish a distinct `task_resumed` event, and the guard preserves a terminal status only for unversioned progress. Providers that publish a monotonic snapshot revision (Grok workflow runs) order by that revision instead, so a resumed run reopens while a stale duplicate is dropped.
- **Background state erased by an admitted progress tick** — confirmed and being fixed on the branch. `claude-task-lifecycle.ts` emits `kind: 'background'` only for the `task_updated` carrying `is_backgrounded`; every later progress event emits `kind: 'subagent'`, and the field-wise merge overwrote it. Reproduced end to end through the real applier: a background row reads `taskKind: subagent`, `isBackgrounded: false` after one progress tick.
- **Task purpose overwritten by the current activity** — confirmed and being fixed on the same commit. Reproduced: a task started as `Find the startup path` renders `Check: checker`, and because the terminal event omits `description` it keeps that value after completion. The running activity stays available through `lastToolName`, so preserving the first description costs no information.

Both remaining fixes are regressions this PR introduced — before it, non-terminal snapshots never reached the merge — so they belong with it rather than in a separate change.

Issue #445 stays open; the superseding pull request will reference it.

PR #359, opened 2026-09-03
feat: tolerate sealed session-history shapes in all readers (phase 2a)

zxch3n:
## What

Reader-side preparation for the upcoming sealed-turn session-history storage. Sealed turns will store tool_call items as **skeletons** `{ type: 'tool_call', kind, status, title?, locations?, ref }` with the execution payload (`content`/`rawInput`/`rawOutput`) kept in the origin machine's local store, and turns will gain an optional derived `summary` and an optional `live` streaming container. **Nothing here changes what any current writer persists** — this PR only makes every reader accept both the old and the new shape.

## Readers that now accept the skeleton / new shape

- **Loro schema validation** (`packages/shared/src/schema.ts`): `historyMessageItemSchema.validate` accepts a `tool_call` with a valid `ref` and no `toolCallId`/`content`; `sessionHistorySchema` declares optional `summary` (`schema.Any`) and `live` (`LoroMap { kind: 'text'|'thought', text: LoroText }`); `sessionExternalHistoryCursorDocSchema` gains `hashVersion`.
- **Shared types** (`packages/shared/src/ai.ts`): `ref?: ToolCallRef` on tool_call; new `ToolCallRef`, `ToolCallPayload`, `TurnSummary`; `ExternalAcpHistorySyncMeta.hashVersion`.
- **Renderer guards** (`packages/components/.../message-content-guards.ts` + new `tool-call-skeleton.ts`): `isMessageContent` accepts skeletons; runtime guards `isToolCallRef`/`isToolCallSkeleton`/`getToolCallStableId` — no casts around the type-required `toolCallId`.
- **Tool-call renderers**: `ToolCallCard` renders skeletons title-only (no output/diff/terminal section) plus an "Execution details are stored on <machine>" placeholder (i18n'd, machine name via `getMachineMetaByIdAtomFamily`) driven by the new `useToolCallPayload(ref)` hook — signature `{ state: 'idle'|'loading'|'ready'|'unavailable', value?: ToolCallPayload }`, stubbed `unavailable` until the Machine RPC task lands. Activity counting in `assistant-turn-render-blocks.ts` classifies from `kind`/`status` alone. `message-copy.ts` (tool calls were already non-copyable → title-only) and `assistant-edited-files.tsx` (derives from the turn-level `fileDiff` evidence store, never from `tool_call.content`) needed no changes — covered by tests.
[...]

PR #460, opened 2026-09-07
fix: unify history writes and optimize streaming updates

zxch3n:
## Summary

One shared HistoryWriter for CLI and renderer, preserving incompatible stored history while validating newly authored data. This is a writer extraction/integration and streaming optimization, not the original unknown-item-only hotfix.

Scope: `35f00a82`, including main through `e12cb225`. **An unresolved queue-activation P1 remains; passing checks are not merge approval.**

## Data flow

```text
CLI / renderer commands → Session facade → HistoryWriter → Loro operations
                                              ↓                ↓
                                    changed-input parsing   Mirror → subscribers
```

- Session Mirror skips whole-state validation; the shared writer preflights local history commands. Other stores retain validation. Unchanged unknown/damaged history is not reparsed or migrated.
- Explicit ACP tool/content JSON extensions remain opaque; malformed known variants cannot use that fallback. Legacy built-in CLI selectors normalize on new writes only.
- Tool fields, stored inputConfig and same-proposal metadata parse independent changes. Fork copies use writer-authenticated raw snapshots, retain target setup rows, and include correlated fork notices.
- Rollback captures only the affected stored range; untouched rows, replacement auto-read acknowledgement and subsequent peer appends survive. Overlapping edits and existing-row structural changes can still reject restoration.
- External imports retain source hashes/ids and record a separate versioned stored-content baseline bound to the document cursor.
- Queue rows remain until history and activation-pointer publication succeed on the normal path; the partial-failure retry gap below is NOT repaired yet.

## Performance and scope limits

Cached schema-derived single-pass parsers, target-local text/thought updates, field-local scalar/fileDiff writes, and a pinned Mirror 2.3.1 single-text-event path-copy patch reduce repeated work. The patch preserves descriptors, old snapshots and notifications; structural/mixed/tree paths retain the original reader. Loro remains 1.15.1; no storage migration.

Recorded at `76d0e9be`: same-writer old/new-reader isolated measurements were 6.067 → 0.180 ms/chunk at 400 entries in Bun and 5.225 → 0.181 ms/chunk at 200 entries in Node. These are NOT whole-app speedups or current-head benchmark reruns. Full-Mirror reads, multi-event tool updates, bulk fork/capture and atomic baseline growth remain concerns. No real 3000-user-round desktop/mobile acceptance or overall 10x claim.

## Validation

- Full local `pnpm check`, formatting and docs checks passed for `35f00a82`.
- After document/test consolidation: 44 focused writer/steer/text-path/forward-compatibility tests and shared typecheck passed. Source-string construction tests were removed; real behavior and compile-failure contracts remain.
- At the preceding main integration: 38 fork/clear/edit-and-resend tests passed using the actual Session facade.
- Earlier real-Loro patched/unpatched comparison passed 3600 deterministic edit steps; actual ACP probe passed 20 single-text chunks. These are historical bounded checks, not deployed-client acceptance.
- Fixtures are synthetic; provider/network/disk boundaries are stubbed where stated in the tests. Local results do not assert current GitHub CI status.

## Review comments and remaining risks

- **Unresolved P1 — queue activation after partial failure:** `appendUserTurn` writes history before `upsertDocMeta(latestUserMsgId)`. If the latter fails, retry sees the existing turn and removes the queue row without repairing its activation pointer. The accepted message can remain undispatched. [Review thread](https://github.com/LodyAI/Lody/pull/460#discussion_r3964211330). This needs a repair that also avoids replaying already-handled queued turns.
- Earlier unknown-field whole-state rejection, missing fork-origin schema and automatic-read rollback invalidation findings have corresponding implementation repairs and regression coverage. This does not resolve the distinct overlapping-rollback risk or automatically resolve review threads.
- `stale_rollback` can still leave restoration incomplete after overlapping/structural edits; the caller logs the error. No durable recovery-copy/crash-recovery workflow is provided.
- Atomic `storedHistoryBaseline` contains all per-turn hashes; no bounded-size hash-chain replacement is implemented. Generic turn reorder remains unsupported. Arbitrary item structural concurrency is not proven safe.

## Relationship to #376

#460 extracted the writer/materializer from #376; #376 should remain open for ConversationView, viewport hydration/rendering and cache ownership. After #460 lands, #376 must integrate it and remove its duplicate writer. That integration has NOT happened yet. Read-path rollback must continue using the shared writer.

#443 primitive/Text hints and #359 skeleton/hash-v2 work remain separate. Future Mirror patches must compose with this patch.

## Documentation

Incremental notes were consolidated into one bilingual owning record, not erased from Git history. [Design and evidence](https://github.com/LodyAI/Lody/blob/fix/session-history-unknown-types/.agents/notes/implemented/architecture/2026-09-07-single-history-writer.md) · [Draft write contract](https://github.com/LodyAI/Lody/blob/fix/session-history-unknown-types/specs/session-history-writes.md).


packages/shared/src/history-write-schema.ts
```diff
@@ -0,0 +1,164 @@
...
+import { ACPSessionConfigSchema, MessageContentSchema, PlanEntrySchema } from './message-schemas';
+
+/** New writes only. Never parse/rewrite the stored history through this schema. */
+export const HistoryEntryWriteSchema = z.object({
+  id: z.string(),
+  role: z.enum(['user', 'assistant', 'system']),
+  timestamp: z.string(),
+  userTurnId: z.string().optional(),
+  acpTurnId: z.string().optional(),
+  items: z.array(MessageContentSchema).optional(),
```
zxch3n, line 16: [P1] Accept the existing `session_fork_origin` notice here. `MessageContent` and `cloneHistoryThroughTurn()` still produce `{ type: 'system_notice', name: 'session_fork_origin', ... }`, but `SystemNoticeSchema` has no matching branch. Both normal and worktree forks later call `targetDoc.updateHistory(() => historyResult.history)`, so this new boundary throws `HistoryWriteError: Invalid history write: items.0: invalid_union`; the target fork is rolled back or marked failed. I reproduced it through `createSessionMirror().historyWriter.append(...)`. Please add the correlated notice schema and a real writer-path fork regression, since the current fork-service mocks pass without exercising this boundary.

zxch3n (commented):
Found one P1 regression at exact head `f0c55094c314697494b48955c06582a7302b0ff6`. GitHub does not allow the PR author's account to submit a Request changes review, so this is recorded as a blocking review comment.

zxch3n (commented):
Re-review at `d3f91aad`: the previous `session_fork_origin` parser failure is fixed, including real writer-path coverage for regular/worktree forks and opaque stored history. One blocking P1 remains: the normal auto-read microtask changes the pending replacement before a failed edit-and-resend commit can roll back, so the rollback receipt becomes stale and the old history tail stays deleted. I reproduced this deterministically with the real history writer plus `attachAutoMarkLatestUserHistoryAsRead`; see the inline thread on `session-edit-and-resend-service.ts:370`.

Issue #451, opened 2026-09-06
[Feature Request] Native pi coding-agent integration over its RPC JSONL mode (first-class, not the third-party pi-acp bridge)


Astro-Han:
### Affected area

Agent runtime / ACP

### Problem or workflow pressure

Pi integration needs a Lody-owned adapter with a defined support boundary, native session identity, and access to Lody’s existing interaction and task UI.

The original proposal suggested embedding a Pi RPC client directly in Lody. The implementation instead uses a standalone ACP adapter, following the same integration boundary as the other adapters.

### Desired outcome

Make Pi usable in Lody through a maintained adapter, while keeping execution and native session files owned by Pi.

### Proposed solution

V1 uses the official Pi CLI in RPC mode:

```mermaid
flowchart LR
    L["Lody: UI and task controls"] -->|ACP| A["Lody-owned Pi adapter"]
    A -->|RPC| P["Official Pi CLI"]
    P --> S["Native Pi session files"]
    P --> T["Packaged question, todo, subagent and MCP tools"]
```

There is no custom SDK worker or separate task store. Lody displays native task state and forwards supported controls.

**V1 implementation: merged**

- [x] Standalone adapter: [LodyAI/acp-extension-pi#1](https://github.com/LodyAI/acp-extension-pi/pull/1).
- [x] Lody task controls and question cleanup: [#605](https://github.com/LodyAI/Lody/pull/605).
- [x] Native session identity and explicit resume, streaming text/thinking/tools, model and thinking configuration, usage reporting, compaction and acknowledged steering.
- [x] Packaged questionnaire, todo and awaited subagent tools. Only the main agent asks the user questions.
- [x] Selected stdio MCP servers through the adapter’s packaged integration.
- [x] Windows process containment and prebuilt native-module packaging checks.

The implementation has been merged; this does not yet mean Pi is available as a released builtin in Lody.

**Release integration: still to confirm or complete**

- [ ] Publish the versioned adapter and managed-runtime artifacts.
- [ ] Register Pi in Lody’s builtin installation and update flow.
- [ ] Verify the shipped integration, including cumulative usage display, installation and upgrade behavior.

**Later candidates: no commitment yet**

These are separate follow-ups to evaluate when needed, not requirements for completing V1:

- Native session-title synchronization without overwriting user-set Lody titles.
- Native history discovery/import and fork support.
- Additional MCP transports.
- Further Pi capabilities where Lody has a corresponding product surface.

### Alternatives or workarounds

[...]

PR #471, opened 2026-09-07
fix(cli): finish stopped steers before releasing session ownership

### LINKED ISSUE #477: [Bug] Stop during steer can block the next user message
### Affected area

Agent runtime / ACP

### Installation method

Built from source

### Lody version or commit

Originally investigated at `d366a5a6454f599f2a7b1f84684909431eed2a12`. The next-turn reproduction also failed on the first #471 revision, `1312e0fc7ce94e6791b2116c371f549c039812b1`.

The fix builds on #571, which keeps the existing turn owner until raw ACP completion or confirmed execution termination. This issue covers the remaining steer waits and handoff during Stop.

### Operating system

macOS 26.6.2 arm64. Reproduced with deterministic executor/client tests; frequency with live providers and behavior on other operating systems have not been established.

### Agent or runtime

Lody's shared acknowledged-steer ACP path. Found during Pi integration (#451), but the reproduction does not depend on a particular provider or model.

### What happened?

Stopping a turn while a steer is pending can leave that steer holding the session operation queue and history-rewrite conflict lease. The interface looks stopped, but the next user message remains pending without reaching the provider.

The outstanding wait can be attachment preparation, configuration application, or the application notification after a steer RPC acknowledgement. Releasing the turn owner alone does not release those waits. A second queued steer must not start preparation after Stop either.

### What did you expect?

- Stop ends local steer waits and lets the queue/lease unwind before interrupting the existing turn owner. An already-applied handoff finishes first.
- The owner remains responsible for outstanding ACP/configuration work until it finishes or the old execution is confirmed terminated. The next input must not overlap a still-running old execution.
- Cancellation of a local wait is not proof of non-delivery. Definite refusal permits requeue; accepted or unknown delivery is not automatically resent by Stop.
- Failure to write a steer outcome must not skip the remaining cancellation cleanup.

### How can we reproduce it?

1. Start turn A through `SessionExecutionService` with a controlled ACP client.
2. Submit steer B and hold preparation, configuration, or its application verdict unresolved. Include the case where the RPC already returned an acknowledgement but no application notification arrives.
3. Stop A, then submit ordinary input C.
4. On the affected implementation, C stays pending without a provider prompt. Releasing only B's outstanding wait allows C to proceed.

The fix should release local waits, keep ownership until old execution completion or confirmed termination, and then let C reach the provider. Also check a late explicit refusal and a second steer queued before Stop.

### How often does it happen?

Reproducible under the controlled conditions above. No live-provider frequency claim.

### Additional context

Proposed fix: #471. This issue is scoped to Stop cleanup and next-input execution, using #571's existing ownership and termination path.

An earlier version of this issue also included a separate recovery problem: a terminal result can be held only in memory when the corresponding user history has not arrived; restarting before history repair loses that exact result. That problem remains unresolved and is deferred from this issue and #471. This fix does not add durable receipts or change restart recovery, and does not promise that missing-history outcomes survive a daemon restart.

### Before submitting

Astro-Han:
## Related issue

Closes #477
Refs #451

## Problem / pressure

Pressing Stop while a steer is pending can leave the session queue and history-rewrite lease occupied. The interface looks stopped, but the next message never reaches the agent.

#571 already keeps the turn owner until the old ACP execution finishes or is confirmed terminated. This fix makes pending steer preparation, configuration and application waits follow that same ownership boundary.

## Summary

- Stop cancels local steer waits before interrupting the owner. An already-applied handoff finishes first.
- The existing owner drains raw ACP and configuration requests, or confirms termination, before releasing the session for the next input.
- The original delivery result survives local cancellation. A late explicit refusal can still requeue the message; accepted or unknown delivery is not resent by Stop.
- Failure to write the steer outcome is logged and does not skip the remaining cleanup.

This PR only covers Stop cleanup and next-message execution. Missing-history results keep the existing in-memory handling; preserving those results across a daemon restart is outside #477 and this PR.

## Visual explanation

Cancelling a local wait does not mean the old execution has ended:

```mermaid
flowchart TD
    A["Stop"] --> B["Cancel local steer wait"]
    B --> C["Finish applied handoff; release queue and lease"]
    C --> D["Existing owner waits for raw ACP/config work or confirmed termination"]
    D --> E["Use delivery result and finish cleanup"]
    E --> F["Release owner; next message can run"]
```

The existing timeout handles an unresponsive execution. If termination fails, the owner continues waiting for the outstanding work.

## Before / after

| Before | After |
| ------ | ----- |
| A pending steer can block the next message after Stop | Local waits and their queue/lease unwind before owner interruption |
| Cancelling a local wait can hide a later refusal | The delivery result remains available to decide whether to requeue |
| Raw configuration or steer work can outlive its local waiter | The owner retains that work until completion or confirmed termination |

## Test plan

- Extended the existing cancellation tests to cover preparation, configuration, application, late refusal and history-write failure, including delivery of the next message after cleanup.
- Controlled AgentClient connections cover cancellation before and after acknowledgement, lease release and later refusal. Attachment and configuration tests check cancellation propagation.
- Removing configuration draining causes premature owner release; ignoring a late refusal incorrectly marks the message failed. Both checks failed as expected, and the restored tests passed.
- Local `pnpm check`, `pnpm format` and documentation checks passed. CI passed on `e4ef2d29`, including desktop smoke. An independent adversarial and simplification review found no remaining blocking issue.
- Live-provider and Windows runtime acceptance were not rerun. The controlled tests do not establish cross-restart recovery guarantees.

- **Review focus:** Trace Stop and natural prompt completion through the steer queue, applied handoff and existing turn finalizer. Check that the next message cannot overlap an unfinished old execution.
- **Decisions to challenge:** Local cancellation must preserve the original delivery result; only definite refusal permits requeue. History-write failure must not skip cleanup.
- **Plausible failures / evidence gaps:** Check late refusal, configuration still running after prompt completion, and failed termination. Live-provider and Windows runtime acceptance remain unverified.

### Authoring context

- **User goal / directives:** Keep #471 focused on Stop blocking the next message, building on #571.
- **Constraints / non-goals:** Reuse the existing owner, queue and termination timeout. Keep history storage, Operation claims and restart recovery unchanged.
- **Risk-bearing decisions:** Applied handoffs finish before interruption. Accepted or unknown delivery is not resent by Stop; a failed outcome write is reported without blocking resource cleanup.
- **Destructive or irreversible behavior:** The existing timeout may terminate an unresponsive execution; earlier tool effects cannot be undone. No session files or user history are deleted.
- **Deliberately not done or tested:** No durable terminal receipts or restart recovery changes. Results held only in memory can still be lost if the daemon restarts before history arrives.
- **Unknowns / confidence:** Tests cover the changed lifecycle paths with controlled responses; their frequency with live providers and Windows runtime behavior have not been rechecked.

### Original user prompt

<details>

````text
可以的，那我们应该 force push 整个重写这个 PR 了。在 571 的基础上面去做改写和重构。包括 PR 正文也需要更新了。
````

</details>

### CODE REVIEW

File: apps/cli/src/agent/agent-client.ts [hunk 1]
```diff
@@ -643,7 +643,7 @@ export class AgentClient implements acp.Client {
   private readonly steerApplicationWaiters = new Map<string, SteerApplicationWaiter>();
   private steerApplicationBarrier: Promise<void> | null = null;
   private activePromptCompletion: ActivePromptCompletion | null = null;
-  private readonly pendingPrompts = new Set<Promise<acp.PromptResponse>>();
```
wibus-wee, line 646: One related design concern: `pendingPrompts` was widened from:

```ts
Set<Promise<acp.PromptResponse>>
```

to:

```ts
Set<Promise<unknown>>
```

so prompt completions and raw steer RPCs can share the same drain.

I don't think `unknown` itself is the bug because the drain only needs settlement, but it seems to expose the underlying issue: the raw steer request is now retained mostly as a “pending/settled” token, while its eventual typed delivery verdict can be lost.

Would it be cleaner to preserve the original typed request and put only a `Promise<void>` settlement wrapper into the cancellation drain? That would keep request draining separate from delivery classification instead of erasing the distinction at this boundary.

wibus-wee (changes requested):
Hi @Astro-Han! I found two correctness gaps, plus one related design concern.

First, I think Stop can currently erase a definitive steer refusal.

After a steer has been submitted, Stop aborts the local application waiter. The raw steer RPC is still retained in the cancellation drain, which is good, but its eventual result is no longer used to classify delivery.

So this sequence appears possible:

```text
steer B submitted
→ Stop aborts local application wait
→ B is recorded as delivery-unknown / failed
→ raw steer RPC later returns invalid request
```

`invalid request` is normally the definitive signal that the provider did not accept the steer, so B should remain eligible for the existing follow-up/requeue behavior. Instead, the late refusal seems to be discarded after the local abort has already won.

I think local cancellation should only release the application wait/lease. The raw RPC should remain the delivery verdict source:

```text
invalid request → definitely refused → requeue
injected ACK    → delivered / do not resend
transport error → unknown / do not resend
```

Second, I think the new durable receipt path still has a write-failure hole.

`setTerminalUserTurnStatus()` currently writes `pendingTerminalUserTurnReceipts` only when the history entry was not found. But if the entry exists and `updateHistory()` itself throws, the function exits before writing a receipt.

That leaves:

```text
history terminal write fails
→ no history outcome
→ no durable receipt
→ restart can still lose the exact terminal outcome
```

It seems safer for the receipt to act as the durable fallback/WAL for the terminal decision, including failures of the initial history write, not only missing-history cases.

Astro-Han:
Thanks for the review, @wibus-wee! I’ve reworked this at `e4ef2d29`. Could you take another look when you have time?

Your late-refusal example helped clarify the boundary: cancelling the local wait must not decide whether the steer was delivered. The original delivery promise now survives cancellation, so a later `invalid request` still permits requeue. The shared request set only handles draining; the delivery verdict is preserved separately.

I also narrowed #477 and this PR to Stop cleanup and next-message execution. The durable receipts, persistence barriers and restart recovery changes are removed. The history-write failure you identified remains a separate recovery problem that I don’t plan to solve in this PR. A failed outcome write does, however, allow the remaining cancellation cleanup to finish.

The existing turn owner remains responsible throughout:

```mermaid
flowchart TD
    A["Stop"] --> B["Cancel local steer wait"]
    B --> C["Finish applied handoff; release queue and lease"]
    C --> D["Existing owner waits for raw ACP/config work or confirmed termination"]
    D --> E["Use delivery verdict and finish cleanup"]
    E --> F["Release owner; next message can run"]
```

Only a definite refusal permits requeue. Accepted or unknown delivery is not resent by Stop.

I’ve run the tests and a separate adversarial review locally, and CI is green, including desktop smoke. I haven’t rerun live-provider or Windows acceptance.

Thanks again for helping catch the ownership issue.

<details>
<summary>中文说明</summary>

谢谢之前指出的问题！这次调整了 Stop 的收尾顺序，也保留了取消后晚到的投递结果，避免把明确拒绝的消息误判成失败。

跨重启的结果保护暂时不打算在这个 PR 里解决，Issue 和 PR 的范围也一起收窄了。我在本地跑了测试和独立的对抗性评审，CI 也通过了。方便时麻烦再帮忙看一下，谢谢！

</details>

apps/cli/src/agent/agent-client.ts
Astro-Han: Thanks, the earlier revision did lose the verdict after local cancellation. At `e4ef2d29`, `SteerPromptRun.delivery` now preserves it independently of the application wait, so a late `invalid request` still reaches the requeue logic.

I’ve kept `pendingPrompts` as a settlement-only collection because it never reads the response. `requestSteeringExtension` validates the response and classifies refusals before `delivery` settles. The finalizer uses that result to decide what happens to the message.

wibus-wee:
Hi @Astro-Han! I think there may still be one ordering race.

If B is waiting for a late refusal and the user sends newer turn C, C can publish `latestUserMsgId = C`. When B is later refused, `requeueUndeliveredSteer(B)` currently writes `latestUserMsgId = B`, which can move the durable dispatch pointer backwards and strand C.

I don't think the finalizer should write the producer-owned `latestUserMsgId` at all. A safer fix would be to mark B back to `pending` and requeue it through the existing dispatch/queue path, while leaving C's `latestUserMsgId` untouched. If that path needs a durable wake-up, use a separate execution-owned requeue marker rather than overwriting the producer pointer.

```text
A active → steer B → Stop → send C → B refused late
→ B and C both run exactly once
→ latestUserMsgId never moves from C back to B
```

Astro-Han:
Thanks, Wibus, for taking the time to review this and work through the races with me. Your feedback caught problems I’d missed.

I’m going to close this PR and take the discussion back to #477. The original Stop issue still needs a fix, but this attempt has reached decisions about message scheduling and state ownership that I’d like to discuss with the maintainers before changing more code.

I’ll leave the commits available for reference. Thanks again for your patience through the revisions.

<details>
<summary>中文说明</summary>

谢谢 Wibus 一直耐心审查，帮我发现了遗漏的竞态问题。我先关闭这个 PR，回到 #477 讨论方案。原来的 Stop 问题还需要修，但消息调度和状态归属的取舍，适合先和维护者讨论清楚，再继续实现。提交会保留供参考。

</details>

PR #479, opened 2026-09-08
feat: honor explicit ACP MCP opt-out

Astro-Han:
## Related issue

Refs #451

## Problem / pressure

Pi's first ACP adapter does not implement MCP, but Lody normally injects its builtin MCP server. Agent configuration probing succeeds while the first prompt fails because Pi rejects the nonempty MCP list. ACP v1 does not provide a stdio opt-out field.

## Summary

Consume the explicit versioned MCP opt-out from LodyAI/acp-extension-core#4 at the shared ACP session-start boundary. Agents declaring no MCP start with no servers; a nonempty frozen workspace selection fails with instructions to clear it. Missing or unrecognized declarations retain existing behavior.

Expose that capability through the existing cache and three composers. Users can see why MCP is unavailable and clear previous selections, including missing catalog entries. This does not implement MCP in Pi or disable Host-owned task workflows.

This change is independent of the Stop/steer lifecycle repair in #471. Pi's declaration is in LodyAI/acp-extension-pi#1. All remain Draft pending combined desktop acceptance.

## Before / after

| Before | After |
| ------ | ----- |
| A no-MCP agent receives builtin MCP and cannot start. | A recognized opt-out starts without builtin or workspace MCP. |
| MCP selection is offered without this capability information. | The menu explains the limitation, prevents additions and lets users clear prior selections. |
| A malformed sibling extension could hide a valid opt-out. | MCP is parsed independently; unknown versions preserve legacy behavior. |

## Test plan

- Focused CLI/shared tests cover startup, replacement, selected-server refusal, parser compatibility and capability persistence.
- Focused component tests cover current/stale capabilities, unavailable MCP and clearing missing selections.
- `pnpm check`, `pnpm format` and `pnpm run docs check` passed. After final simplification, 142 affected CLI tests and CLI typecheck passed.
- A local integration branch combining this change and #471 built successfully with `pnpm e2e:build`. Real Electron acceptance is running; this Draft does not claim it has passed.
- The MCP Storybook state was added; the standalone Storybook bundle reached chunk rendering and exceeded the default 4 GB heap. The actual Electron production build passed.

- **Review focus:** Trace the frozen MCP ids into `AgentClient.buildMcpServers`, then the live capability through cache persistence and all three composers.
- **Decisions to challenge:** The shared explicit opt-out preserves ACP defaults, rejects requested servers before fallback, and leaves existing next-start selection semantics intact.
- **Plausible failures / evidence gaps:** Check unknown catalog ids, stale cached capabilities and restored/replacement sessions; combined Electron acceptance is pending.

### Authoring context

- **User goal / directives:** Make the first Pi adapter usable without MCP and state that limitation clearly, with independent review and behavior-based validation.
- **Constraints / non-goals:** No Pi MCP implementation, ACP v2 migration, agent-name exception, release or changes to Stop/steer ownership in this PR.
- **Risk-bearing decisions:** Live initialization is the runtime authority; cached support only informs UI. Only exact version-one false opts out, preserving legacy agents.
- **Destructive or irreversible behavior:** No catalog deletion or automatic selection rewrite. Users explicitly clear selections; the cache version bump requests fresh discovery.
- **Deliberately not done or tested:** Windows packaging and managed distribution belong to the adapter release work. Full desktop acceptance is not yet claimed.
- **Unknowns / confidence:** Focused behavior checks support the boundary; final independent review and combined desktop checks will determine readiness. The Core submodule commit is publicly fetchable before merge.


Astro-Han:
Closing because we have chosen to implement MCP support in the Pi adapter instead of shipping Pi with MCP disabled.

This removes the need for the opt-out startup, capability-cache, and composer changes proposed here. Lody already provides a stdio path for its built-in MCP tools, so the adapter can use the existing contract. The corresponding Core proposal is also closed: https://github.com/LodyAI/acp-extension-core/pull/4.

The replacement will be developed in https://github.com/LodyAI/acp-extension-pi/pull/1. An isolated bridge experiment passed, but the implementation and full Desktop acceptance are still pending. The separate Stop/steer and informational-result fixes are unaffected by this decision.

PR #506, opened 2026-09-08
fix(components): render blob images inline in tool call resource blocks

### LINKED ISSUE #462: [Bug] 点击reading media，展开了媒体的base64
### Affected area

Desktop app

### Installation method

Desktop release

### Lody version or commit

0.92.0

### Operating system

macOS arm64

### Agent or runtime

kimi 0.39.1

### What happened?

点击reading media，显示了base64

Image

### What did you expect?

应该显示图片吧

### How can we reproduce it?

让ai阅读图片，点击reading media

### How often does it happen?

Every time

### Relevant log output

```shell

```

### Additional context

_No response_

### Before submitting

clanzhang:
## Related issue

Closes #462

## Problem / pressure

When an ACP agent reads an image (e.g. Kimi's "reading media" tool), the tool returns a `resource` content block with binary `blob` data and an `image/*` mimeType. The `StandardToolContentBlock` component in `view.tsx` only checks for `text` in the resource — it falls through to rendering a "Download blob" link using `content.resource.uri` as the href. The URI is the resource's original identifier (e.g. `file:///path/to/image.png`), not a downloadable URL, so clicking it exposes the raw base64 string to the user instead of rendering the image.

## Summary

Add a `blob` branch to the `resource` case in `StandardToolContentBlock`. When the resource contains a `blob` with an `image/*` mimeType, construct a data URL via the existing `buildSafeBase64DataUrl()` helper and render an inline `` tag, matching the pattern already used by the `image` content type case. Non-image blobs fall through to the existing download link.

## Before / after

| Before | After |
| ------ | ----- |
| Tool returns a `resource` with `blob` + `image/*` mimeType → "Download blob" link → raw base64 displayed to user | Tool returns a `resource` with `blob` + `image/*` mimeType → inline `` renders the image |

## Test plan

- Have an agent (e.g. Kimi) read an image via its tool
- Observe the tool output in chat — the image should render inline instead of showing base64 data
- Verify non-image blob resources (e.g. PDF) still show the download link
- Verify text resources still render as markdown
- Run `pnpm --dir packages/components test` if available

- **Review focus:** `packages/components/src/components/ai-gui/view.tsx` — the `resource` case in `StandardToolContentBlock` (around line 6121). Verify the new `blob` branch correctly reuses `buildSafeBase64DataUrl()` and the `image/*` mimeType guard.
- **Decisions to challenge:** Whether non-image blob types should also get inline rendering (audio, PDF). The current fix is image-only to match the reported bug; extending to other types is a separate scope.
- **Plausible failures / evidence gaps:** The fix depends on `content.resource` having `blob` and `mimeType` fields from the ACP SDK `EmbeddedResource` type. If a specific agent sends blob data under a different key (e.g. `data` instead of `blob`), it would not be caught. The `buildSafeBase64DataUrl` helper validates both fields so malformed data renders nothing rather than breaking.

### Authoring context

- **User goal / directives:** Fix issue #462 so that image resources from agent tool calls render visually instead of exposing base64 data.
- **Constraints / non-goals:** Only fix the rendering path in `StandardToolContentBlock`. Do not change the ACP protocol, CLI-side materialization, or other content block types.
- **Risk-bearing decisions:** The `buildSafeBase64DataUrl` helper already validates mimeType format and base64 alphabet, so no new sanitization was added. The data URL is consumed only by an `` tag's `src` attribute.
- **Destructive or irreversible behavior:** None. The change is purely additive — a new branch before the existing fallback.
- **Deliberately not done or tested:** Audio and PDF blob rendering. The issue specifically concerns images. Other blob types continue to show the download link. Integration testing against a live agent was not performed.
- **Unknowns / confidence:** High confidence for image/* blobs. The pattern is a direct copy of the existing `image` case at line 6067-6081. Moderate uncertainty on whether all agents use `blob` as the key name — the ACP SDK `EmbeddedResource` type distinguishes `TextResourceContents` (with `text`) from `BlobResourceContents` (with `blob`), so this matches the protocol contract.

### DIFF
--- a/packages/components/src/components/ai-gui/view.tsx
+++ b/packages/components/src/components/ai-gui/view.tsx
@@ -6128,6 +6128,24 @@ const StandardToolContentBlock = ({
           />
         );
       }
+      // Handle blob resources (binary data like images)
+      if ('blob' in content.resource) {
+        const src = buildSafeBase64DataUrl(content.resource.mimeType, content.resource.blob);
+        if (!src) return null;
+        // Render image blobs inline, other blobs as download links
+         if (content.resource.mimeType?.startsWith('image/')) {
+          return (
+            <div className="space-y-2">
+              <div className="text-xs font-medium text-muted-foreground">Resource</div>
+              <img 
+                src={src}
+                alt={content.resource.uri || 'Resource image'}
+                className="max-h-80 w-full rounded-lg object-contain"
+              />
+            </div>
+          )
+         }
+      }
       const href = sanitizeToolContentHref(content.resource.uri);
       if (!href) return null;
       return (
--- a/packages/components/src/components/ai-gui/view.tsx
+++ b/packages/components/src/components/ai-gui/view.tsx
@@ -6128,23 +6128,23 @@ const StandardToolContentBlock = ({
           />
         );
       }
-      // Handle blob resources (binary data like images)
-      if ('blob' in content.resource) {
+      // Render image blobs inline; fall through to download link for everything else.
+      if (
+        'blob' in content.resource && 
+        typeof content.resource.mimeType === 'string' &&
+        content.resource.mimeType.startsWith('image/')
+      ) {
         const src = buildSafeBase64DataUrl(content.resource.mimeType, content.resource.blob);
-        if (!src) return null;
-        // Render image blobs inline, other blobs as download links
-         if (content.resource.mimeType?.startsWith('image/')) {
-          return (
-            <div className="space-y-2">
-              <div className="text-xs font-medium text-muted-foreground">Resource</div>
-              <img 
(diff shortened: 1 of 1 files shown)


wibus-wee (changes requested):
I traced the actual Kimi path. Lody pins `acp-extension-kimi@38fb45e`, where structured tool results are serialized into ACP text via `JSON.stringify(...)`, not emitted as `resource.blob`.

So this blob renderer doesn't address the actual base64 path. Also, the new `if (src)` branch is missing a `return`, so it currently renders nothing anyway.

I think the better fix is in `acp-extension-kimi`: normalize Kimi's media result into standard ACP image/resource content, then let the existing UI renderer handle it.

PR #542, opened 2026-09-09
fix(terminal): resolve colors without reduced-motion transitions

### LINKED ISSUE #541: [Bug] Reduced motion makes embedded terminal text match the background
### Affected area

Desktop app

### Installation method

Desktop release

### Lody version or commit

Observed in desktop 0.92.1. Reproduced against the public renderer source at
`4400744a37e0d64b2d21a15e15cf96b5f2486a21` using Chromium.

### Operating system

macOS 26.6.2 arm64. Source regression reproduced in Playwright Chromium 145.0.7632.6.

### What happened?

With the system's Reduce motion preference enabled, the embedded terminal can
look completely blank: foreground, cursor, and ANSI colors resolve to the same
color as the background. The shell continues to produce output.

The public `buildTerminalTheme` function reuses a hidden span, assigns each
`--terminal-*` token to its `color`, and immediately reads computed style. The
global reduced-motion rule gives every element a nonzero `transition-duration`
(`0.01ms !important`). Because the probe retains `transition-property: all`,
successive reads see the beginning of a color transition rather than the target.

### What did you expect?

The terminal should resolve each requested theme color correctly with Reduce
motion enabled, including when it is first opened or its theme is changed.

### How can we reproduce it?

1. On macOS, enable Accessibility > Motion > Reduce motion.
2. Start Lody and open the embedded terminal in a local session with a dark theme.
3. Run a command that produces text and ANSI colors, such as
   `printf '\033[31mred\033[0m \033[32mgreen\033[0m\n'`.
4. Observe that the terminal may appear blank despite the shell producing output.

A source-level reproduction loads the real shared stylesheet and
`packages/components/src/components/terminal/terminal-theme.ts` into Chromium,
emulates `reducedMotion: 'reduce'`, sets distinct background/foreground/ANSI
custom properties on a host, and calls `buildTerminalTheme(host)` synchronously.
With a `#101010` background, foreground/red/green all resolve to `rgb(16, 16, 16)`.
The same test passes under `no-preference`.

### How often does it happen?

Every time

### Additional context

Setting `probe.style.transitionProperty = 'none'` before the probe is appended
fixes the source regression in both motion modes. The same resolver is used
for terminal initialization and theme updates, so this can remain a local fix
without changing global accessibility styles or adding asynchronous waits.
The browser check also covers a dark-to-light palette change, cursor color,
alpha selection color, and removal of the disposable probe.

### Before submitting

nzhl:
## Related issue

Closes #541

## Problem / pressure

With Reduce motion enabled, the embedded terminal can appear blank because its
foreground, cursor, and ANSI colors resolve to its background color. The global
`0.01ms` transition duration affects the hidden span used by `buildTerminalTheme`,
so immediate computed-style reads can return the start of a color transition.

## Summary

Disable transitions on the disposable color probe before appending it. This
keeps terminal creation and theme updates synchronous and leaves the global
reduced-motion behavior intact. Add a Chromium regression using the real
stylesheet and resolver, plus a short decision note.

## Visual explanation

```diff
 const probe = document.createElement('span');
 // Existing positioning, visibility, and pointer-event styles.
+probe.style.transitionProperty = 'none';
 host.appendChild(probe);
 // Assign each color token, then read getComputedStyle(probe).color immediately.
```

## Before / after

| Before | After |
| ------ | ----- |
| Reduced motion can make foreground, cursor, ANSI colors, and selection match the background. | Each token resolves to its requested color, including alpha selection colors. |
| Normal motion resolves the palette correctly. | Normal motion retains the same palette, and subsequent palette changes resolve correctly. |

## Test plan

- Chromium regression command: `VITE_PREVIEW_PUBLIC_BASE_DOMAIN=example.invalid pnpm --filter @lody/components exec playwright test tests/e2e/terminal-theme.spec.ts --workers=1 --reporter=line`.
- Before the source fix: the reduced-motion case failed and the normal-motion case passed. After the fix: both cases passed.
- Both cases verify dark and light palettes on the same host, foreground/cursor/red/green colors, selection alpha, and probe cleanup; they load the actual shared stylesheet and resolver without mocking computed style or waiting on timers.
- `pnpm check` passed with Node 22.22.2, including typechecks, lint, existing tests, and boundary guards; the component suite passed 445 files / 3306 tests.
- `pnpm format`, formatting checks for the new test and note, `pnpm run docs check`, and `git diff --check` passed. Unrelated formatter output was excluded, preserving the resolver's existing formatting.
- The packaged desktop app and non-Chromium engines were not rebuilt or tested; the bug and patch were exercised directly in Chromium using the public renderer source.

- **Review focus:** Check the probe setup in `terminal-theme.ts` and the browser regression's use of the real reduced-motion stylesheet, since both initial and later terminal themes use this resolver.
- **Decisions to challenge:** Disabling only the probe's transition property preserves the global accessibility rule and synchronous API without timers.
- **Plausible failures / evidence gaps:** The Playwright test covers Chromium rendering, not a packaged Electron lifecycle, and belongs to the existing opt-in `test:e2e` suite rather than root Vitest CI.

### Authoring context

- **User goal / directives:** Contribute a permanent fix for an embedded terminal that appears blank when the system Reduce motion preference is enabled.
- **Constraints / non-goals:** Keep the patch local to color resolution; leave global accessibility preferences, terminal fonts, shell processes, and theme tokens unchanged.
- **Risk-bearing decisions:** Exclude the disposable probe from CSS transitions before attachment so sequential computed-style reads return target colors; retain the existing cleanup and theme-update paths.
- **Destructive or irreversible behavior:** The code affects a temporary DOM element and adds no persistent-data operations; reverting the commit restores the previous resolver behavior.
- **Deliberately not done or tested:** No desktop release build or cross-engine run; browser verification targets the CSS transition behavior with the real public source and synthetic color tokens.
- **Unknowns / confidence:** High confidence in the reproduced color-resolution fix from the failing-before/passing-after browser test; packaging and full desktop lifecycle remain outside that evidence.

### DIFF
--- a/.agents/notes/proposed/bug-fix/2026-09-09-terminal-reduced-motion-colors.md
+++ b/.agents/notes/proposed/bug-fix/2026-09-09-terminal-reduced-motion-colors.md
@@ -1,0 +1,41 @@
+# Resolve terminal colors without transitions
+
+Status: proposed
+Translation: pending
+
+## Abstract
+
+With reduced motion enabled, the terminal's synchronous CSS color probe can
+return the background color for its foreground, cursor, and ANSI palette,
+making a working terminal appear blank. This patch disables transitions on
+the disposable probe before it enters the document. Chromium reproduces the
+failure on the unchanged resolver and passes with the patch in both motion
+modes, including a dark-to-light palette change. The global accessibility
+styles and terminal lifecycle remain unchanged.
+
+## Decision
+
+`buildTerminalTheme` owns CSS-to-xterm color resolution for both terminal
+creation and theme updates. The global reduced-motion rule sets a nonzero
+`transition-duration` on every element; the probe's default transition property
+is `all`. Each immediate computed-style read can therefore observe the start
+of a color transition instead of the requested token.
+
+Set `transitionProperty` to `none` on the probe before appending it. Changing
+the global reduced-motion rule would affect unrelated components, while
+waiting for a transition would make a synchronous resolver asynchronous.
+
+## Evidence and limits
+
+The Playwright regression uses the real stylesheet and resolver through the
+existing Storybook server, with synthetic CSS tokens. It checks foreground,
+cursor, red/green ANSI colors, alpha selection color, a palette change, and
+probe cleanup under `reduce` and `no-preference`.
+
+Before the patch, only the reduced-motion case fails; after the patch, both
+cases pass. This is Chromium renderer verification, not a rebuilt desktop
+release or verification of every terminal renderer and operating system.
+
+`pnpm check` passes with Node 22.22.2, including the existing component suite.
+`pnpm format` and `pnpm run docs check` pass; unrelated formatter output was
+excluded from the patch, preserving the resolver's existing formatting.
--- a/packages/components/src/components/terminal/terminal-theme.ts
+++ b/packages/components/src/components/terminal/terminal-theme.ts
@@ -50,6 +50,8 @@ export function buildTerminalTheme(host: HTMLElement): ITheme {
   probe.style.position = 'absolute';
   probe.style.visibility = 'hidden';
   probe.style.pointerEvents = 'none';
+  // Global reduced-motion styles must not animate synchronous color reads.
+  probe.style.transitionProperty = 'none';
   host.appendChild(probe);
   try {
     const theme: ITheme = {};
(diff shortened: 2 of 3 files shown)


Leeeon233:
Thanks!

PR #564, opened 2026-09-10
feat: refresh subscription quota before sessions

### LINKED ISSUE #547: [Feature Request] Refresh subscription quota on app startup and manual provider refresh
### Affected area

Desktop app

Users need current subscription quota information before starting a conversation because the remaining quota influences which model and reasoning effort they should choose for the next task.

In Lody Desktop 0.92.1 on Windows, Codex quota can remain unchanged while no conversation is running. The usage popover can even display a reset time that has already passed. Starting a conversation then updates the displayed quota, but by that point the user has already selected a model and reasoning effort.

The refresh button beside Codex in Settings -> Agents also does not provide an explicit quota refresh: its current path refreshes ACP capabilities. A successful refresh therefore does not mean the displayed subscription quota is current.

### Desired outcome

Users should be able to make an informed model and reasoning-effort choice before sending a prompt:

1. When opening the desktop app and the relevant machine/provider becomes available, retrieve current quota information for configured subscription providers that support it, without requiring a conversation to run first.
2. When manually clicking the existing refresh button beside a provider in Settings -> Agents, refresh its quota information as well as its capabilities.
3. Update the quota shown in both provider settings and the conversation composer from the refreshed result, including usage/remaining percentage, applicable windows, and reset times.
4. If fresh quota cannot be retrieved, make the unavailable or stale state visible instead of presenting an expired cached value as current. Do not assume an expired window means zero usage.

### Proposed solution

Use the provider's supported quota-query mechanism from the machine-side runtime on application startup/availability and explicit provider refresh. Persist and distribute the resulting quota snapshot through the existing quota data path.

Keep these queries independent of submitting a user prompt. The exact implementation is left to maintainers; the important behavior is that opening the app or explicitly refreshing supplies useful quota information before the next conversation starts.

- Starting a conversation updates quota, but requires choosing the model and reasoning effort before seeing the information needed for that choice.
- Checking quota in another client works, but interrupts the workflow and leaves Lody's own display stale.
- The existing provider refresh button updates capabilities without explicitly updating persisted quota.

### Trade-offs or non-goals

- This request concerns quota visibility in the open-source desktop app and its machine-side runtime, not hosted billing or account management.
- Support providers with an available quota-query capability; do not require unsupported providers to fabricate quota values.
- Startup refresh should not block opening the app or starting a conversation. Deduplicate concurrent queries and avoid repeated requests for unavailable machines.
- Continuous polling, automatic model selection, account switching, and automatic retry of quota-blocked turns are outside this request.

### Additional context

The motivating observation was a Codex usage popover showing an unchanged percentage and a reset time eight days in the past, then updating after a conversation started.

Source inspection of the installed 0.92.1 bundle and public source found:

- Codex publishes quota on session setup/auth refresh and on `account/rateLimits/updated` notifications: [CodexAcpServer.ts](https://github.com/LodyAI/acp-extension-codex/blob/f9dbc8c0e98bc11ad13e0f487ecb2b6e0672cfea/src/CodexAcpServer.ts#L836), [CodexEventHandler.ts](https://github.com/LodyAI/acp-extension-codex/blob/f9dbc8c0e98bc11ad13e0f487ecb2b6e0672cfea/src/CodexEventHandler.ts#L410).
- The usage popover consumes supplied quota data and formats reset timestamps: [session-usage-popover.tsx](https://github.com/LodyAI/Lody/blob/4400744a37e0d64b2d21a15e15cf96b5f2486a21/packages/components/src/components/sessions/session-usage-popover.tsx#L58).
- The manual capability probe does not register the quota-update callback, so it does not explicitly persist quota returned during the probe.

Related but distinct: #361 requests automatic retry after quota reset; this request is about obtaining current quota before the user chooses how to start a conversation.

### Before submitting

Dante-dan:
## Related issue

Closes #547

## Problem / pressure

Subscription quota currently appears only after a live conversation emits usage data. This leaves Settings and the composer without a reliable pre-conversation quota snapshot and can make an expired cached percentage look current.

## Summary

- Query advertised ACP subscription quota through the existing startup and manual capability probe, with a bounded five-second quota deadline.
- Persist fresh/stale metadata beside machine-owned quota cache rows while preserving legacy rows and preventing older probes from overwriting newer live updates.
- Retry a startup pass when machine availability changes, while deduplicating concurrent config refreshes.
- Keep cached percentages visible with a stale label, and show an explicit unavailable state when an eligible provider has no snapshot.
- Add English and Simplified Chinese copy plus a behavior spec and decision note.

## Visual explanation

```mermaid
flowchart LR
  A[Startup availability] --> P[Shared capability probe]
  B[Settings refresh] --> P
  P --> Q{Quota query advertised?}
  Q -->|yes| C[Bounded Core query]
  Q -->|no| N[Notification fallback]
  C --> M[Machine Flock snapshot]
  N --> M
  L[Live quota update] --> M
  M --> U[Settings and composer]
  M --> S[Preserve value + mark stale on failure or elapsed reset]
```

## Before / after

| Before | After |
| ------ | ----- |
| Quota appears only after conversation activity and expired values can look current. | Startup and manual provider refreshes populate quota before a prompt; cached values remain visible with honest stale/unavailable states. |

## Test plan

- `pnpm --filter @lody/components exec vitest run tests/session-usage.test.ts tests/session-usage-popover.test.tsx tests/provider-row-reauthentication.test.tsx tests/create-workspace-runtime-meta-recovery.test.ts --no-file-parallelism`
- `pnpm --filter lody exec vitest run src/agent/acp-capabilities.test.ts src/lib/loro/machine-document-capabilities.test.ts tests/session-execution-service.test.ts --no-file-parallelism`
- `pnpm --filter @lody/shared run typecheck`
- `pnpm --filter @lody/components run typecheck`
- `pnpm --filter lody run typecheck`
- `pnpm format`
- `pnpm run docs check`
- `pnpm check:quick`

- **Review focus:** Trace quota ownership from `acp-capabilities.ts` through `MachineDocument.writeRateLimits` into startup scheduling and the two UI surfaces.
- **Decisions to challenge:** Check the five-second optional query deadline, provider partitioning, and the rule that any elapsed reset marks the retained snapshot stale.
- **Plausible failures / evidence gaps:** Pay particular attention to same-millisecond live/query ordering and availability changes during an active startup pass; both have focused regression coverage.

### Authoring context

- **User goal / directives:** Make subscription quota available before a conversation and keep stale or unavailable states honest instead of presenting a fabricated zero.
- **Constraints / non-goals:** Reuse the existing ACP capability probe and machine/provider cache; continuous polling, account switching, automatic model selection, and retrying quota-blocked turns remain outside this feature.
- **Risk-bearing decisions:** Query failures preserve cached percentages as stale, and attempt-start timestamps prevent slow probe results from replacing newer live data.
- **Destructive or irreversible behavior:** The change performs no deletion or migration; legacy quota rows are normalized in place on refresh and cached values remain recoverable after failures.
- **Deliberately not done or tested:** The implementation stays within the existing built-in provider partition and does not add a cloud billing path or a second background transport.
- **Unknowns / confidence:** Capability-gated provider behavior remains adapter-dependent; local probe, persistence, race, migration, startup, rendering, type, lint, documentation, and boundary checks pass.


Leeeon233:
Hi @Dante-dan, I'm adding a new interface to acp-extension-core. The quota will automatically refresh when the relevant component is mounted.  Thanks for this PR, but I have to turn it off first.

PR #591, opened 2026-09-11
feat: add Bub as a builtin ACP provider

Leeeon233:
## Summary

Adds **Bub** to the Agent Config **Built-in** provider list. Lody launches the user-installed `bub acp serve` command and speaks ACP to it.

## Behavior

- `resolveBuiltinACPProcessLaunch` returns `command: "bub"`, `args: ["acp", "serve", ...extra]`. `bub` resolves from the same augmented login-shell PATH as other local ACP agents.
- Lody does not download, version, or sign in to Bub. Creation requires a live probe, so a missing command surfaces an install hint with a link to https://bub.build/docs/tutorials/acp-server/ instead of failing on the first turn. The same link appears in the provider row failure tooltip.
- Capability source version is the static `builtin-bub:local`; an explicit refresh re-probes after an upgrade.
- Static capabilities are empty on purpose, so modes, models, and config options come from the real probe.
- Bub pushes no session title, so Lody keeps its isolated title generator.
- The icon is a `currentColor` trace of the wordmark in https://github.com/bubbuild/bub (the repository ships raster logos only).
- Registers `bub` in the skill-directory map and the local session-control builtin literal. That literal was missing `grok`; both are added so it matches `BUILTIN_AGENTS`.

## Notes

- The request said "builtin list", so Bub is in the **Built-in** group rather than the ACP registry. That group normally means Lody-managed runtimes; Bub is user-installed, so the row carries no `Registry`/`Custom` badge even though Lody only owns the integration.
- No install-state detection is added, per the request.

## Testing

- `@lody/shared`: typecheck plus the full suite (1125 tests), including the new `ai-bub` contract tests.
- `lody` CLI: typecheck plus the `agent-setting`, `provider-setup-manager`, and `acp-capabilities` suites.
- `@lody/components`: typecheck plus the `agent-config-dialog`, `onboarding-flow`, `acp-selector-options`, `provider-row-reauthentication`, and `onboarding-summary-agent` suites.
- `pnpm lint:fast`, `pnpm format:check`, `pnpm lint:i18n`, docs `status`/`check`, and all three boundary guards.
- Not verified: a real end-to-end ACP session against a machine with Bub installed. The probe-failure prompt is covered by the dialog test, not a live runtime.

Agent note: `.agents/notes/implemented/feature/2026-09-11-builtin-bub-provider.md`.


frostming:
While the old command still works, we prefer `bub acp`, and the docs have been updated accordingly. Can you update the command as well?

frostming:
BTW bub now has a one-inline installation command as well:
```
curl -fsSL https://bub.build/install.sh | bash -- --preset acp
```

`--preset acp` will install all necessary dependencies needed by acp

PR #643, opened 2026-09-12
fix(cli): author agent commits with the GitHub no-reply address

zxch3n:
## Related issue

None. No existing Issue covers this. #521 is related but insufficient — it only addresses the owner machine's gitconfig, not which email the requester resolves to.

## Problem / pressure

`Session.updateGitIdentity` exports the resolved requester identity as `GIT_AUTHOR_*`/`GIT_COMMITTER_*`, and those variables override `~/.gitconfig`. Lody resolved the stored account email, which for an account with GitHub's **Keep my email addresses private** plus **Block command line pushes that expose my email** is exactly the address GitHub refuses: every push of an agent commit fails with `GH007: Your push would publish a private email address`. The user cannot ship the work the agent just did, and the address they chose to keep private ends up in the commit object.

The obvious fix — always prefer the account's GitHub no-reply address — is wrong in the other direction: it ignores where the session actually pushes. A repository on GitLab, on a GitHub Enterprise host, or with no remote at all would commit with a `users.noreply.github.com` address that attributes nothing there and can be rejected.

## Summary

Use `<id>+<login>@users.noreply.github.com` as the requester's commit email **only when the session workdir pushes to github.com**.

- `git-identity.ts` owns the decision, because `resolveSessionGitIdentity` already has the session `cwd`. It reads the workdir remotes, lets `origin` decide alone when it exists (a GitLab repo with a GitHub mirror is still a GitLab repo), prefers each remote's push URL, and counts only `github.com` / `gist.github.com` as GitHub — GitHub Enterprise is deliberately excluded.
- `session-user-resolver.ts` cannot make this call: it is cached per user across sessions and never sees a workdir. It now reports `githubNoreplyEmail` **beside** the account email instead of replacing it, so the non-GitHub fallback order keeps its own meaning.
- The turn carries the address as the optional `userGitHubNoreplyEmail` request field, frozen with the rest of the requester identity rather than re-read at execution time.

Resulting order: on a github.com remote the no-reply address wins (the machine identity is exactly what GH007 rejects for these accounts); everywhere else the existing policy is unchanged — machine owner prefers the repository/machine Git identity, non-owners never read it, then the account email, then `LodyAI <agent@lody.ai>`.

Deliberately unchanged: `GIT_*` is still always exported (dropping it when host gitconfig is set would attribute a teammate's turn on a shared daemon to the machine owner); worktree archive backup identity; no new user-facing setting.

## Visual explanation

Simple change: one new branch at the top of `resolveSessionGitIdentity`, plus an optional field threaded beside the existing `userName`/`userEmail`. The full order is written out in the Summary, the Spec, and the Note.

## Before / after

| Before | After |
| ------ | ----- |
| github.com origin, private account email → account email → push rejected with GH007 | → `4324+ada@users.noreply.github.com`, push accepted, attributed to the same account |
| gitlab.com origin, GitHub-connected account, gitconfig set → account email (overrides gitconfig) | → gitconfig `user.email`, no github.com address on a GitLab commit |
| No remote / GitHub Enterprise origin | → gitconfig, then account email, then `agent@lody.ai` (unchanged) |
| GitHub origin, incomplete GitHub profile | → existing fallback (unchanged) |
| Teammate's turn on a shared daemon | Never the machine owner's identity (unchanged) |

## Test plan

- `apps/cli`: `pnpm test tests/git-identity.test.ts tests/session-user-resolver.test.ts` → 43 passed. Covers https and ssh GitHub origins, gitlab.com, no remote, a GitHub mirror behind a GitLab origin, a GitHub Enterprise host, an incomplete GitHub profile, the non-owner and neutral fallbacks, remote-URL host parsing, and `git remote -v` parsing with push-URL precedence. Three cases run against real temporary git repositories so the `git remote -v` / `git config` reads are actually exercised, with inherited `GIT_AUTHOR_*`/`GIT_COMMITTER_*` stubbed out.
- `apps/cli`: `pnpm test` → **2737 passed, 1 failed, 4 skipped (262 files)**. The failure is `tests/worktree-gc.test.ts`, a macOS `/private/var` vs `/var` tmpdir realpath mismatch; verified pre-existing by re-running it against `HEAD` without these changes.
- `packages/shared`: `pnpm test` → 1127 passed.
- Root `pnpm typecheck` clean; `pnpm lint` 0 errors; `pnpm check:public-boundary` and `pnpm check:platform-boundaries` pass; `pnpm run docs check` → 0 errors.
- Not executed: an end-to-end push from an account with private-email push protection. GH007 is taken from GitHub's push protection behavior, not reproduced in CI.

Docs updated in the same PR: `apps/cli/src/session/AGENTS.md`, `apps/cli/src/session/README.md`, `specs/git-commit-identity.md` + `.zh.md` (still `draft`), and the bilingual Agent Note `.agents/notes/implemented/bug-fix/2026-09-12-github-noreply-commit-email.md`.

**Risk: medium · Confidence: high** — `risk:medium`. The blast radius is every agent commit's author/committer email, and this adds one optional field to `SessionCreateRequest`/`SessionChatRequest`; it is additive and only ever set by daemon-internal producers, so no client sends it to an older strict validator. Known limitation: hosts such as `ssh.github.com` are not recognised as GitHub, which costs those users the fix rather than breaking them.

### DIFF
--- a/.agents/notes/implemented/bug-fix/2026-09-12-github-noreply-commit-email.md
+++ b/.agents/notes/implemented/bug-fix/2026-09-12-github-noreply-commit-email.md
@@ -1,0 +1,57 @@
+# Prefer the GitHub no-reply address for the requester's commit email
+
+Status: implemented
+Translation: current
+
+[中文](2026-09-12-github-noreply-commit-email.zh.md)
+
+## Abstract
+
+A session turn exports the requester's resolved commit identity as `GIT_AUTHOR_*`/`GIT_COMMITTER_*`,
+and those variables override `~/.gitconfig`, so whatever Lody resolves is what the agent's commits
+carry. The resolver preferred the stored account email and fell back to the GitHub no-reply address
+only for missing-email placeholders; for an account that enables "Keep my email addresses private"
+together with "Block command line pushes that expose my email", that real address makes GitHub
+reject every push of an agent commit with GH007, leaving the user unable to ship the work. The fix
+inverts that single preference: whenever the profile carries both a GitHub account id and login, the
+requester's commit email is `<id>+<login>@users.noreply.github.com`, and the account email is used
+only when no such pair exists. The trade-off is that a user whose account email is public and who
+expected it on the commit object now sees the no-reply address instead; GitHub attributes both to
+the same account, so attribution is unchanged.
+
+## Decision
+
+The GitHub no-reply address is the strictly better commit email whenever it exists. It attributes
+the commit to the same GitHub account that later opens the pull request, it is never rejected by
+the private-email push block, and it does not publish an address the account chose to keep private.
+The stored account email carries no capability the no-reply lacks, so the previous preference only
+ever traded a working push for a cosmetically nicer address.
+
+The change is confined to `SessionUserResolver.toSessionUserProfile`. The downstream ownership
+policy in `git-identity.ts` is untouched:
+
+- machine owner: effective machine/repository Git identity, then this resolved identity, then the
+  neutral `LodyAI <agent@lody.ai>`;
+- non-owner requester: this resolved identity, then the neutral identity.
+
+Two alternatives were rejected. Skipping the `GIT_*` export entirely when the host `user.email` is
+set would let a remote or multi-user daemon attribute a teammate's turn to the machine owner, which
+is exactly what the ownership policy exists to prevent. Adding a user-facing setting was rejected as
+out of proportion: the no-reply address is correct for every account that has one, so there is
+nothing left to configure.
+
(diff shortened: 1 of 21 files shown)


zxch3n (commented):
Approve. No blocking findings. The diff now chooses <id>+<login>@users.noreply.github.com when both usable GitHub fields exist, correctly falls back for incomplete profiles, and leaves owner/non-owner selection plus unconditional GIT_* injection unchanged. Suggestion: rerun the focused Vitest files in a dependency-complete checkout; this worktree lacks node_modules (vitest: command not found). pnpm run docs check reaches the script but reports only pre-existing unrelated broken links; the changed session AGENTS.md remains under 8192 bytes.

PR #645, opened 2026-09-12
fix(cli): accept an unsplit ACP terminal command line [risk:medium]

### LINKED ISSUE #469: [Feature Request] Compatibility for ACP agents sending an unsplit shell command line getting failure with ENOENT (-2)
### Affected area

Agent runtime / ACP

When an ACP agent (e.g. codebuddy-code) calls `createTerminal` with a full, un-split shell command line in `command` (e.g. `ls -la /foo`) and an empty `args`, lody spawns `command` directly as an executable. The OS cannot resolve an executable literally named `ls -la /foo`, so the call fails with `ENOENT` (errno 2, surfaced to the agent as `-2`). Spec-conformant ACP clients (Claude Code, Codex, Zed) split the line into `command` + `args[]` and work fine, so the failure only appears with agents that send an un-split line.

### Desired outcome

Lody should execute such terminal commands successfully (interoperating with agents that do not pre-split their command line), while continuing to behave exactly as before for spec-conformant callers.

### Proposed solution

I plan to open a PR that adds compatibility handling in `ShellTerminalManager.startProcess` (the `createTerminal` execution path): when `command` contains whitespace and `args` is empty, run it through a shell (`/bin/sh -c` on POSIX, `cmd.exe /c` on Windows). A plain executable name with split `args` is spawned unchanged.

Today the only workaround is to have the agent route shell execution through other tools such as Monitor, but that tool is not 100% reliable, so the failure mode above remains in practice.

### Trade-offs or non-goals

Splitting the command into `command` + `args[]` is the agent's responsibility per the Agent Client Protocol, not lody's; the compatibility handling is implemented purely for interoperability, not because lody is at fault.

Non-goals: this does not fix the agent's behavior, and it does not broaden support to other degenerate shapes (e.g. an empty `command` with the line passed in `args`).

### Additional context

_No response_

### Before submitting

zxch3n:
Risk: 🟡 medium | Confidence: high — the shell fallback is gated on three conditions and covered by tests, but no live agent that sends an unsplit line was exercised, and the Windows branch is verified only through an injected platform.

## Related issue

Closes #469

## Problem / pressure

ACP `terminal/create` carries the executable in `command` and its argv in `args`. Several agents instead send the whole shell line in `command` with empty `args` (a bare `ls -al`, or a relayed `bash -lc …`). Lody spawned that literally, so the OS had no such executable, and the failure surfaced differently per sandbox:

- Linux: the cgroup sandbox awaits the child's pid, so `spawn` rejected and the agent received an untyped JSON-RPC error carrying errno `-2`. Several agents drop the ACP session over it.
- darwin: the fallback sandbox resolves the handle before the OS reports the failure, so `terminal/create` returned a terminal id and `terminal/wait_for_exit` never resolved — the turn hung.

`ShellTerminalManager`'s `onError` hook only logged, so no spawn failure ever completed a waiter on either platform.

## Summary

- `resolveTerminalSpawnTarget` rebuilds a command only when `args` is empty, `command` contains whitespace, and `stat` finds no file at that path relative to the workdir: `sh -c` on POSIX, `cmd.exe /c` on Windows. Non-login and non-interactive, so no user profile runs.
- Split `command` + `args` keeps going straight to `SessionSandbox.spawn`; `args.length > 0` is never wrapped, and an executable path that contains a space is still spawned directly.
- `waitForSpawnConfirmation` awaits the child's pid when it is not already set, so a spawn failure rejects `createTerminal` instead of publishing a terminal that can never exit. Node assigns `pid` synchronously on success, so a started process does not wait.
- `ENOENT`/`EACCES` become `TerminalSpawnError`, which `AgentClient.createTerminal` answers as `acp.RequestError.invalidParams` (`-32602`) instead of a bare errno.
- The terminal `onError` hook now records an exit status, so a process error after a successful start completes pending `wait_for_exit` calls.
- `resolveCustomACPSetting` expands a leading `~` in a custom agent's launch command (human-typed, not expanded by `spawn`). Terminal command lines are deliberately excluded.

## Visual explanation

```mermaid
flowchart TD
  A["terminal/create(command, args)"] --> B{"args.length > 0?"}
  B -- yes --> S["spawn(command, args)"]
  B -- no --> C{"whitespace in command?"}
  C -- no --> S
  C -- yes --> D{"stat(workdir/command) is a file?"}
  D -- yes --> S
  D -- no --> E["spawn(sh -c | cmd.exe /c, line)"]
  S --> F{"pid assigned?"}
  E --> F
  F -- yes --> G["terminalId"]
  F -- "no: ENOENT / EACCES" --> H["RequestError.invalidParams (-32602)"]
  G -.-> I["process error later"] --> J["record exit status, release waiters"]
```

## Before / after

| Before | After |
| ------ | ----- |
| `command: "ls -al"`, `args: []` spawned `ls -al` as `argv[0]` and failed. | The same call runs through `sh -c` (`cmd.exe /c` on Windows) and returns output. |
| `command: "git"`, `args: ["status"]` spawned directly. | Unchanged: split argv never goes through a shell. |
| `command: "/tmp/my tool"`, `args: []` spawned directly. | Unchanged: an existing file is spawned directly, never quoted into a shell. |
| A spawn `ENOENT` reached the agent as an untyped errno `-2`, or (darwin) as a live terminal id whose `wait_for_exit` hung. | `terminal/create` rejects with JSON-RPC `-32602`, and any late process error records an exit status. |
| A custom agent launched with `~/bin/agent` failed to start with `ENOENT`. | The leading `~` is expanded before spawn. |

## Test plan

- `vitest run tests/terminal-manager.test.ts tests/agent-setting.test.ts`: 44 passed (13 + 31). Against a stub sandbox: the unsplit line on POSIX and Windows, untouched split argv, an executable path containing a space, a command line that already carries args, both spawn-failure shapes (sandbox rejection and late error channel), a pending wait completed by a process error, and the `-32602` mapping at the ACP boundary. Against a real process (`createNoopSessionSandbox`, POSIX only): an unsplit `printf` line returns its output with exit code 0, and an unresolvable executable rejects instead of publishing a terminal.
- Ablation: removing the `stat` guard fails the space-in-path case, removing the spawn confirmation fails the late-error case, and removing the `onError` exit record leaves the wait hanging until its timeout.
- `pnpm check`: typecheck, lint, i18n, and boundary checks passed; 2722 of 2727 CLI tests passed. The one failure, `tests/worktree-gc.test.ts`, fails identically on the unmodified base commit `702b5aa1` — a local macOS `os.tmpdir()` `/var` vs `/private/var` mismatch, unrelated to this change.
- `pnpm format` and `pnpm run docs check` (no errors).
- Not exercised: a live agent that sends an unsplit line, and real `cmd.exe` behavior on Windows (that branch is covered through an injected platform, and the real-process cases skip on Windows).

- **Review focus:** `resolveTerminalSpawnTarget` and `waitForSpawnConfirmation` in `apps/cli/src/session/terminal-manager.ts`, plus the relaxed `handleExit` registration guard.
- **Decisions to challenge:** whether the `stat`-is-a-file check is the right line between an unsplit command line and a space-containing executable path, and whether `invalidParams` (rather than `internalError`) is the correct code for a command that cannot be executed.
- **Plausible failures / evidence gaps:** a sandbox handle that neither assigns a pid nor emits `spawn`/`error` would now block `createTerminal` where it previously returned; real Windows `cmd.exe` quoting is untested on a Windows host.

### Authoring context

- **User goal / directives:** fix #469 surgically — keep the spec path as the default, add the narrow shell fallback, and never let a spawn failure hang `wait_for_exit`.
- **Constraints / non-goals:** no `/bin/bash -lc`, no wrapping when `args` is non-empty, no shell rebuilding for `terminal/create` lines beyond the documented case, and no change to agent-side command splitting.
- **Risk-bearing decisions:** the shell fallback executes an agent-supplied line under `sh -c`, where metacharacters apply; it is reachable only for a line the agent already intended as a shell command, and the `stat` guard keeps real executable paths out of it.
- **Destructive or irreversible behavior:** none; no migration, no persisted state, no cleanup path changed.
- **Deliberately not done or tested:** `~` is not expanded for terminal command lines (agent-supplied, and the shell already expands it when it runs), and PR #470 was not merged as-is because it lacks the file check and both failure paths.
- **Unknowns / confidence:** high confidence in the gated fallback and the typed rejection; residual risk is live-agent behavior and Windows `cmd.exe` quoting.

### DIFF
--- a/.agents/notes/implemented/bug-fix/2026-09-12-acp-terminal-unsplit-command-line.md
+++ b/.agents/notes/implemented/bug-fix/2026-09-12-acp-terminal-unsplit-command-line.md
@@ -95,4 +95,5 @@ Not verified here: a live agent that sends an unsplit line (the reporter's
 branch is covered only through the injected platform, and the real-process cases
 skip on Windows.
 
-Issue: [#469](https://github.com/LodyAI/Lody/issues/469).
+PR: [#645](https://github.com/LodyAI/Lody/pull/645), closing
+[#469](https://github.com/LodyAI/Lody/issues/469).
--- a/.agents/notes/implemented/bug-fix/2026-09-12-acp-terminal-unsplit-command-line.md
+++ b/.agents/notes/implemented/bug-fix/2026-09-12-acp-terminal-unsplit-command-line.md
@@ -1,0 +1,98 @@
+# Accept an unsplit ACP terminal command line
+
+Status: implemented
+Translation: pending
+
+## Abstract
+
+ACP `terminal/create` carries an executable plus argv, but several agents send
+the whole shell line in `command` with empty `args`; spawned literally it fails
+with `ENOENT`, and the failure was invisible or fatal depending on the sandbox —
+Linux surfaced an untyped errno `-2` that made agents drop the session, darwin
+returned a terminal id whose `wait_for_exit` never resolved. Lody now spawns the
+protocol pair as before and rebuilds a command only in that one shape (empty
+`args`, whitespace, no file at that path), through a non-interactive, non-login
+`sh -c` (`cmd.exe /c` on Windows). A spawn that still fails rejects
+`terminal/create` with a JSON-RPC code and records an exit status, so no waiter
+hangs. The fallback cannot tell a genuine one-word-with-spaces intent from an
+unsplit line beyond the file check, so it stays deliberately narrow rather than
+becoming a general shell mode.
+
+## Problem and ownership
+
+Splitting a command line is the agent's job under the Agent Client Protocol, and
+`ShellTerminalManager` owns process execution. The compatibility handling is
+therefore interoperability at the execution boundary, not a correction of Lody's
+protocol reading: `command` + `args` stays the contract and the default path.
+
+Two distinct defects met in the same call. The first is resolution: cross-spawn
+gets `ls -al` as `argv[0]` and the OS has no such executable. The second is
+reporting. `sandbox.spawn` behaves differently per platform — the cgroup sandbox
+awaits the child's pid and rejects, while the POSIX fallback resolves the handle
+and lets the error arrive later on the process error channel. The terminal's
+`onError` hook only logged, so the late error resolved nothing: `createTerminal`
+had already returned an id and `waitForTerminalExit` waited forever. Both
+platforms therefore needed the same typed answer.
+
+## Decision
+
+Three changes, each narrow:
+
+1. `resolveTerminalSpawnTarget` rebuilds a command only when `args` is empty,
+   `command` contains whitespace, and `stat` finds no file at that path relative
+   to the workdir. The file check is what preserves an executable path that
+   contains a space; `args.length > 0` is never wrapped, so a caller that
+   already split its argv keeps exact spawn semantics.
+2. `waitForSpawnConfirmation` awaits the child's pid when it is not already set,
+   so a spawn failure rejects `createTerminal` instead of publishing a terminal
+   that can never exit. A started process costs nothing here: Node assigns `pid`
+   synchronously, so the check returns immediately.
+3. `ENOENT`/`EACCES` become `TerminalSpawnError`, which `AgentClient.createTerminal`
+   answers as `acp.RequestError.invalidParams` (`-32602`). A command that cannot
+   be executed is a bad request, and a numeric JSON-RPC code is what lets an
+   agent report the failure instead of treating it as a dead connection. The
+   terminal's `onError` hook now also records an exit status, so a process error
+   after a successful start completes pending waits rather than hanging them.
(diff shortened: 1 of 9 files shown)


zxch3n (commented):
Approved in substance: I found no blocking P0/P1 issues in the diff. The change preserves direct command,args spawning, limits shell fallback to empty args + whitespace + non-file path, avoids bash -lc, converts ENOENT/EACCES to numeric -32602, and completes waiters on process errors. Added tests cover POSIX/Windows routing, space-containing paths, split args, spawn failures, waiter completion, and custom-agent-only ~ expansion. Targeted tests could not run because this worktree has no installed dependencies (vitest: command not found).

PR #653, opened 2026-09-12
fix(cli,shared): render engine-opened agent turns as their own history entries

### LINKED ISSUE #640: [Bug] Engine-opened agent turns (cron fires, task wakes) merge into the previous user turn instead of rendering as their own turn
### Affected area

Agent runtime / ACP

### Installation method

Built from source

### Lody version or commit

2acd5117 (current `main`; also observed on recent release builds)

### Operating system

macOS arm64 (reported; platform-independent by analysis)

### Agent or runtime

Kimi Code managed runtime (builtin `kimi` agent, ACP), observed with runtime 0.39.1-lody on model `kimi-code/k3`

### What happened?

When the Kimi engine opens a turn of its own — a cron fire, a task wake — its output is appended to the **previous user turn's assistant entry** instead of appearing as its own turn. In the session timeline this shows up as:

- The user turn's real reply is folded into the collapsed "thinking process" (思考过程) area.
- The engine turn's text becomes the visible "answer" of the user turn it interrupted-adjacent, which is a non-sequitur reply to the user's question.
- After a daemon restart, the same class of updates is dropped entirely (logged only at debug level).

Root cause, verified against a live session's engine wire log and the Lody daemon logs:

1. The kimi ACP server deliberately forwards content events for **every** turn, including turns the engine opened itself (by design, so background replies are not swallowed).
2. Those updates carry **no turn identity** (`_meta.lody.turnId` is only minted for user-visible turns as the fork position).
3. Lody's history layer routes updates without a client-turn target to the last finalized turn's assistant entry (`lateACPUpdateTarget`, which intentionally never expires by time, to catch stragglers of the finished turn). Engine-turn updates are indistinguishable from such stragglers, so the two turns merge into one entry.
4. With no active/finalized target in memory (e.g. after a daemon restart), the updates hit the drop path.

Two additional symptoms of the same gap, seen in the same session:

- The daemon's idle GC SIGKILLed the agent process in the middle of an engine-opened cron turn, because no client turn was tracked as active.
- The user's turn right before the cron fire died to a provider 500 with zero output; the user then resent the message several times, hitting queued-turn cancellation and `another turn is already in progress` errors because the engine turn hogging the agent is invisible in the UI.

### What did you expect?

Turns are displayed one-to-one with the engine's own turn model:

- The user-triggered turn stands as its own turn with its real reply, even if it ended empty or failed.
- The engine-opened turn (cron fire / task wake) renders as a separate, following turn with its own work and output, and is marked finished when it ends — not merged into the user turn, and not dropped when no client turn is in memory.

### How can we reproduce it?

1. In a Lody session on the builtin Kimi agent, let the agent schedule a one-shot cron checkpoint (CronCreate) a few minutes out.
2. Send a user message so a user turn is running (or just finished) when the cron fires.
3. Wait for the cron to fire.
4. Observe: the cron turn's thinking/tool calls/text are appended into the user turn's assistant entry; the visible reply of that user turn becomes the cron turn's text.
5. Variant: restart the daemon while no client turn is active and let a cron fire — its output is dropped with only a debug-level `Dropping ACP update without an active/finalized assistant entry target` log line.

### How often does it happen?

Every time (whenever an engine-opened turn fires adjacent to a user turn or with no client turn in memory).

### Relevant log output

```shell
# engine wire (kimi session): user turn 40 ends, 1s later the cron fire opens engine turn 41
turn.ended  turnId=40 reason=completed
turn.prompt origin=cron_job   # <cron-fire ...> envelope, prompt redacted
turn.steer  origin=cron_job
# Lody history afterwards: the assistant entry keyed to the USER turn (assistant:<userTurnId>)
# ends with engine turn 41's text as its visible reply.
```

### Additional context

Analysis and fix direction (already implemented on a branch, PRs to follow):

- The kimi ACP server stamps every update of an engine-opened turn with a non-numeric `_meta.lody.turnId` (`auto:<engineTurnId>` — can never parse back into a `session/fork` position) plus `_meta.lody.turnOrigin`, and emits a `_meta.lody.turnEnded` marker when the turn ends.
- `@lody/shared` history-apply gives only those marked turns their own `assistant:autonomous-<turnId>` entry; every other stamped id keeps the exact legacy last-wins restamping (claude's per-message boundary uuids and codex collab child turns are boundaries inside one client turn).
- `apps/cli` routes marked updates to a synthesized autonomous target even while a client turn owns the session (also covers rich content), instead of dropping them when no target exists.
- The fix spans this repository and `LodyAI/acp-extension-kimi`; the kimi side takes effect with the next managed-runtime artifact, the Lody side is safe to ship first (inert for unstamped runtimes).

Known follow-ups not covered by the fix: idle GC killing an agent process mid-engine-turn; replay/import not classifying cron fires (a reloaded session can still surface cron-fire XML as ordinary user messages); no origin label on autonomous turns in the UI yet.

### Before submitting

tommy0103:
## Related issue

Closes #640

> **Companion PR: #654 must land together with this one.** It carries the
> lifecycle guards the codex review rounds identified (fork/edit-resend
> suppression of `auto:` entries, ACP `exit`-path marker cleanup, engine-turn
> Stop routing, replay-window bypass, monitor-state coverage). The split
> exists only to keep this PR under the community 1000-line gate; review them
> as one change and merge back-to-back. (An origin-divider UI was tried and
> reverted — it breaks the stream index contract; tracked in #660.)

## Problem / pressure

The Kimi engine opens turns behind the client's back — a cron fire, a task wake — and its ACP server forwards their content by design. Those updates carried no turn identity, so the history layer routed them into the last client turn's assistant entry (the never-expiring finalized-turn target built for stragglers), or dropped them when no client turn existed in memory. Observed in production sessions: the user turn's real reply folded into a collapsed "thinking process" while the cron turn's text became the visible answer; a session showing "completed" (`executionState: idle`) while an engine turn ran for 15+ minutes; the 20-min idle GC SIGKILLing the agent process mid-cron-turn; engine-turn output dropped after a daemon restart.

## Summary

- **Wire identity (kimi ACP server, PR LodyAI/acp-extension-kimi#9, submodule pin bump included):** engine-opened turns are stamped `_meta.lody.turnId = auto:<n>` (non-numeric, can never parse into a `session/fork` position) + `_meta.lody.turnOrigin`; a `_meta.lody.turnEnded` marker is emitted when such a turn ends.
- **Turn-aware grouping (`@lody/shared` history-apply):** only `auto:`/`turnOrigin`-marked updates get turn-aware routing — owner entry → adopt only a delta-accepting entry → else create `assistant:autonomous-<turnId>`. Every other stamped id keeps the exact legacy last-wins restamping (claude per-message boundary uuids and codex collab child turns are boundaries inside one client turn). `turnEnded` finalizes the owning entry once. `acpTurnOrigin` is persisted on the entry.
- **Routing (`apps/cli`):** marked updates always take a synthesized autonomous target — never the active/finalized client turn's — covering the no-target drop and rich content. An engine-turn activity marker (set on first marked update, cleared on end marker or ACP process termination — process-bounded, no wall clock) feeds `hasActiveTurn` (idle-GC guard) and upgrades live status `unknown` → `running`.
- **UI:** the chat stream renders an eyebrow divider ("Scheduled task" / "Background task") before engine-opened turns, since they have no user message.

## Visual explanation

```mermaid
flowchart LR
  subgraph kimi [Kimi engine + ACP server]
    UT[user turn 40] --> T41[cron turn 41, origin=cron_job]
    T41 --> S["_meta.lody: turnId=auto:41, turnOrigin, turnEnded"]
  end
  S --> R{apps/cli enqueueACPUpdate}
  R -->|marked| AT[autonomous target<br/>assistant:autonomous-auto:41]
  R -->|unmarked| LT[active/finalized client target]
  AT --> HA{history-apply}
  LT --> HA
  HA -->|auto:/turnOrigin only| E1[own entry: adopt-if-accepting<br/>or create autonomous entry]
  HA -->|any other id| E2[legacy last-wins restamping<br/>claude uuids, codex collab]
  S --> M[engineTurn activity marker<br/>transient store]
  M -->|set| BUSY[live status running + GC guard]
  M -->|turnEnded / process exit| IDLE[clear]
  E1 --> UI[divider: Scheduled task<br/>+ own finished turn]
```

## Before / after

| Before | After |
| ------ | ----- |
| Engine turn output appended to the previous user turn's entry; its text became the visible "answer", real reply folded | User turn keeps its reply; engine turn renders as its own following turn |
| Engine-turn updates dropped after daemon restart | Routed to `assistant:autonomous-<turnId>`; content survives |
| Engine turn entry never finalized (perpetual streaming) | `turnEnded` marker finalizes it once |
| `executionState: idle` while the engine turn runs; idle GC kills the agent mid-turn | Live status reads `running`; GC guard sees the engine turn; marker clears on end/process exit |
| No visual cue that a turn came from a cron fire | Eyebrow divider labels the origin |

## Test plan

- `packages/shared`: full suite 1133 passing, incl. new cases — engine-turn separation, cross-batch continuity, died-before-output adoption hole, end-marker finalization (once-only timing), claude multi-uuid last-wins, codex collab inline restamping.
- `apps/cli`: full suite 2715 passing (3 failures reproduce on unmodified `origin/main` in this environment: two Claude credential-store probes, one macOS `/var` vs `/private/var` worktree-GC assertion) — new store cases for autonomous-target gating and engine-turn activity note/clear/replace.
- `packages/components`: 16 passing, incl. divider emission on both incremental and cache-reuse paths. `acp-server` (submodule): 163 passing (1 pre-existing env e2e failure on base).
- tsc clean in shared/cli/components/acp-server; prettier, oxlint, `check-i18n`, `pnpm run docs check` clean.
- End-to-end on OSS desktop + locally built stamped runtime: cron turn after a user turn rendered as its own finished turn with origin divider; session presence read `running` during the engine turn, `idle` after; provider-500-empty-turn + adjacent cron fire no longer merges.
- Skipped: replay/import consistency and cron-prompt surfacing (follow-ups in the issue); GC-kill repro used reasoning + logs, not a 20-min live wait.

- **Review focus:** `packages/shared/src/acp/history-apply.ts` (the three-way turnId branch in `apply()` + `ensureEntryForAcpTurn` adoption condition + `finalizeEntryOnce`), `apps/cli/src/lib/session-transient-store.ts` (`autonomousACPUpdateTargetFrom` gating, engine-turn marker lifecycle), `apps/cli/src/lib/message-handler.ts` (`enqueueACPUpdate` target precedence), kimi `acp-server/src/session.ts` (stamp + end-marker emission).
- **Decisions to challenge:** gating all new behavior on `auto:`/`turnOrigin` markers only (legacy path kept for every other stamped id); adoption allowed only on delta-accepting entries; engine-turn liveness bounded by ACP process termination rather than a wall clock; `auto:<n>` as the non-fork id format.
- **Plausible failures / evidence gaps:** task-wake turns now also render separately (subagent UX change); a GC-killed engine turn never emits its end marker (unfinished entry, same as any crashed turn); replay/import does not classify cron fires (reload can surface cron XML as user messages — known follow-up); kimi stamping only lands with the next managed-runtime artifact, and the client side is untested against that artifact in production.

### Authoring context

- **User goal / directives:** Investigate and fix a session display bug where a user-message turn interrupted-adjacent to a cronjob fire collapses into a "thinking process"; user then directed: two turns should render separately, keep the blast radius kimi-only, verify end-to-end, file the issue and open the PRs.
- **Constraints / non-goals:** No behavior change for claude/codex/pi/grok/deepseek or unstamped runtimes; fork-at-turn semantics untouched; no replay/import rework, no cron-prompt surfacing, no GC/lifecycle redesign beyond the marker-based guard.
- **Risk-bearing decisions:** new history field `acpTurnOrigin` (additive Loro schema change); deterministic entry ids `assistant:autonomous-<turnId>`; non-numeric `auto:` ids can never resolve into `session/fork` positions; legacy last-wins restamping preserved byte-for-byte for all unmarked ids.
- **Destructive or irreversible behavior:** none — history writes are additive; a duplicate end marker cannot move an entry's terminal timing; failed flush groups requeue with at-least-once semantics unchanged.
- **Deliberately not done or tested:** cron-prompt user-message surfacing, replay/import origin classification, UI label for the cron prompt text, live GC-kill repro (20-min wait); each is recorded as a follow-up in #640.
- **Unknowns / confidence:** high confidence in unit/replay-level correctness and the single-machine e2e; moderate on multi-turn subagent UX (wake turns separating) and on behavior before the stamped artifact ships (inert by design, asserted by tests but not observed in production).

### Original user prompt

<details>

````text
现在似乎有个 bug，在 fusion-model 下有个 session，你会发现如果当 user message trigger 的一个轮次被 cronjob 打断，那么这个轮次会被折叠为思考过程，这个显示显然不是我们期望的，你去 check 一下
````

</details>

### DIFF
--- a/.agents/notes/implemented/bug-fix/2026-09-12-engine-turn-entries.md
+++ b/.agents/notes/implemented/bug-fix/2026-09-12-engine-turn-entries.md
@@ -1,0 +1,128 @@
+# Engine-opened agent turns get their own history entries
+
+Status: implemented
+Translation: current
+
+[中文](2026-09-12-engine-turn-entries.zh.md)
+
+## Abstract
+
+An agent engine can open turns behind the client's back — a kimi cron fire, a
+task wake — and its ACP server deliberately forwards their content. Those
+updates carried no turn identity, so Lody routed them to the last client turn's
+assistant entry: the user watched their real reply get folded into a collapsed
+thinking block while the cron turn's status text became the visible answer,
+after a daemon restart such updates were dropped outright, and the session
+showed "completed" while the engine turn was still running (its process could
+even be reaped by the idle GC). The fix threads a turn identity through the
+wire (`_meta.lody.turnId` = `auto:<n>` plus `_meta.lody.turnOrigin` for
+engine-opened turns) and routes updates by that identity: an engine-opened
+turn renders as its own entry, reports the session busy until its end marker,
+while boundary ids that live inside one client turn keep their legacy
+rendering. The remaining limit is deployment: the kimi stamping ships with
+the next managed-runtime artifact, while the Lody-side layers ship with Lody
+and degrade to the old behavior for unstamped agents.
+
+## The non-obvious constraint
+
+Late-update routing exists for a legitimate reason: a turn's own output can
+settle after `session/prompt` returns, so the finalized turn stays the routing
+target with no wall-clock expiry. The same rule is what swallowed engine-opened
+turns — from the routing layer's view, a cron turn's first chunk is
+indistinguishable from a straggler of the turn that just ended. Any fix
+therefore needs a real turn identity on the wire; boundary guesses based on
+timing or prompt-flight state were considered and rejected as the primary
+mechanism because consecutive engine turns and text-only engine turns are
+indistinguishable without one (a cli-only heuristic remains a viable fallback
+for unstamped runtimes, trading exactness for deployability).
+
+## Ownership
+
+- The kimi ACP server (`acp-extension-kimi` submodule) stamps every update of a
+  turn the engine opened itself with `auto:<engineTurnId>` (non-numeric, so it
+  can never parse back into a fork position) plus the origin kind. User-visible
+  turns keep plain fork positions and carry no origin marker.
+- `history-apply` in `@lody/shared` owns grouping, scoped to kimi-stamped
+  turns only: an `auto:` id or `turnOrigin` marker (engine-opened) adopts the
+  current entry only while it still accepts deltas — a finalized-but-unstamped
+  entry left by a turn that died before producing output must not adopt an
+  interrupting turn — or creates `assistant:autonomous-<turnId>`; every other
+  id keeps the exact legacy last-wins restamping, because claude's per-message
+  boundary uuids and codex collab child turns label boundaries inside one
+  client turn and must not splinter its rendering.
+- `apps/cli` routes origin-stamped (or `auto:`-prefixed) updates to a
+  synthesized autonomous target even when an active/finalized target exists:
+  an engine-opened turn can never belong to the client turn's entry, and
+  routing early also covers rich content, which bypasses turn-aware regrouping
+  at write time.
+- The kimi ACP server emits `_meta.lody.turnEnded` when an engine-opened turn
+  ends, and history apply finalizes the owning entry (once — a duplicate
+  marker must not move the terminal timing). Without the marker the entry
+  would render as perpetually streaming, because `message.finished` is the
+  renderer's only streaming verdict and finalization otherwise happens only
+  for client-dispatched turns. The origin is also persisted as
+  `entry.acpTurnOrigin` so the UI can label such turns.
+- `apps/cli` also maintains an engine-turn activity marker in the transient
+  store: set on the first stamped update of an engine-opened turn, cleared on
+  its end marker or on ACP process termination (liveness is process-bounded,
+  never wall-clock — a turn whose marker is lost releases the session as soon
+  as its process dies). `hasActiveTurn` reads it so the idle GC cannot reap
+  the agent process mid-engine-turn, and the live-status RPC upgrades
+  `unknown` to `running` so the session stops showing "completed" while the
+  engine turn is still working.
+
+## Alternatives and trade-offs
+
+- Cli-only boundary detection (seal the finalized turn after the finalize drain
+  and route everything later to an autonomous entry) covers the main case with
+  zero cross-repo coordination, but cannot split consecutive engine turns,
+  loses the boundary across daemon restarts, and hard-codes a kimi special
+  case. Kept as a possible fallback, not implemented.
+- Stamping plain engine turn numbers as fork positions was rejected: Lody
+  parses `session/fork` targets back out of `turnId`, so engine turns must be
+  non-numeric.
+- Grouping every stamped id (the first revision of this fix) was rejected on
+  review: it splintered claude's per-message boundary uuids into empty entries
+  and stole the fork-relevant `acpTurnId` from the real reply entry, and it
+  pulled codex collab child output out of the parent turn's inline rendering.
+- A generic owner-routing tier for ordinary stamped ids (route a late chunk of
+  turn N back to turn N's entry) was dropped to keep the blast radius
+  kimi-only: a straggler that crosses into the next turn keeps its pre-existing
+  merge behavior rather than gaining a new code path.
+- Task-wake turns now render as their own entries too, separating a held
+  subagent's report from the user turn that prompted it. This is consistent
+  with the engine's own turn model but is a visible change worth validating in
+  the subagent UX.
+- Left as follow-ups: replay/import does not classify cron fires, so a
+  reloaded session can still surface cron-fire XML as ordinary user messages;
+  the UI does not yet label autonomous turns by origin or show the cron
+  turn's trigger prompt.
+
+## Evidence and verification
+
+- Reproduced from the production session wire (`~/.kimi-code/.../wire.jsonl`):
+  a cron turn firing one second after a user turn completed had its final text
+  persisted as the visible reply of that user turn's assistant entry.
+- New unit coverage: engine-turn separation, cross-batch continuity, the
+  died-before-output adoption hole, end-marker finalization (and its
+  once-only terminal timing), claude multi-uuid last-wins, and codex collab
+  inline restamping in `@lody/shared`; autonomous-target gating (origin and
+  `auto:` prefix) and engine-turn activity note/clear/replace in `apps/cli`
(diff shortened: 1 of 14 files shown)


tommy0103:
Thanks — valid catch. Autonomous entries (`auto:` ids) are now suppressed in all four selection paths: the CLI edit-and-resend walk (`session-edit-and-resend-service.ts`), `cloneHistoryThroughTurn` (rejects an engine-opened fork source), the renderer edit-and-resend enablement (`session-chat-interface.tsx`), and the renderer fork gate (`view.tsx`). Regression tests cover both the backward walk skipping an autonomous entry (boundary resolves to the preceding real provider turn) and the fork-source rejection.

The fix landed in the stacked follow-up #654 (commit d27b19e2) rather than this PR to keep #653 under the community 1000-line gate; behavior and tests are identical to having it here.

tommy0103:
Correct on the facts, and this sequencing is deliberate (also called out in this PR's Summary and in the linked decision note):

- The manifest pin can only move **after** LodyAI/acp-extension-kimi#9 merges: the checksummed artifact must be built from the upstream-merged submodule revision, not from the fork commit this branch currently pins for development (`1c5acbc6`). Artifact production/publication to the R2 channel is an external distribution step I cannot perform from here.
- Shipping this PR before that artifact lands is safe by construction: every new behavior is gated on the `auto:`/`turnOrigin` markers, which an unstamped runtime never emits, so installed users see byte-identical behavior (unit-tested: unstamped updates follow the exact legacy path). The bug simply stays unfixed for them until the artifact arrives — nothing breaks.

Follow-ups required after #9 merges, before this PR's behavior reaches users: (1) build + publish the new managed-runtime artifact, (2) bump `kimi-runtime-manifest.json` (version/sha256/sourceCommit), (3) repoint the submodule pin here to the upstream merge commit. Happy to do (2) and (3) in this PR or a fast follow-up, whichever the maintainer prefers.

PR #654, opened 2026-09-12
fix(cli): lifecycle guards for engine-opened turns (fork/resend, exit cleanup, Stop routing, replay window)

tommy0103:
## Related issue

Refs #640

> **Stacking note: this PR is stacked on #653 and the two must land together.**
> It carries the lifecycle guards from the codex review rounds (fork/edit-resend
> suppression of `auto:` entries, ACP `exit`-path marker cleanup, engine-turn
> Stop routing + failed-stop semantics, replay-window bypass, monitor-state
> coverage). The rest of the diff shown here is #653's content; the policy
> check's 1000-line finding measures the stacked diff and clears once #653
> merges and this branch is rebased. Review as one change; intended order:
> **#653 first**, then this immediately after.

## Problem / pressure

#653 separates engine-opened turns into their own history entries. That new entry type then flowed through existing lifecycle predicates that were written for client-dispatched turns only: fork and edit-and-resend treated a deliberately non-fork-position `auto:` id as a fork boundary; the engine-turn activity marker leaked when the ACP process died via `exit` (never `terminated`); Stop matched no client turn and reported success while the work continued, and a failed stop would have cleared state anyway; a live engine turn firing during the `loadSession` replay-suppression window was dropped as replay; the machine monitor and the disconnected-session idle reset could not see the turn.

## Summary

- Fork/edit-resend hardening: autonomous (`auto:`) entries are skipped while walking to the real provider boundary in both the CLI and renderer selection paths, `cloneHistoryThroughTurn` rejects an engine-opened fork source, and the renderer no longer offers Fork on such entries.
- Lifecycle: the engine-turn activity marker also clears on the ACP process `exit` event; machine monitor and `resetMachineDisconnectedSessionsToIdle` read the marker.
- Stop routing: a stop request matching no client turn cancels the active engine-opened turn through plain ACP `session/cancel`; the marker clears only after a delivered cancel, and a failed or undeliverable cancel keeps it and returns `success: false`.
- Replay window: `enqueueACPUpdate` resolves the autonomous identity before replay suppression, so a live engine turn firing during the `loadSession` window is never dropped as replay (replay carries no autonomous marker).

(The origin-divider UI that briefly lived here was **reverted**: a separate stream item breaks the `history index == chat items index` coordinate contract that `handleScrollToMessage` and search navigation rely on — see #660 for the properly constrained design.)

## Visual explanation

Simple change: predicate-level guards and one cancel-path branch, each mirrored by unit tests; no control-flow restructuring.

## Before / after

| Before | After |
| ------ | ----- |
| Fork/edit-resend forwards a non-position `auto:` id; fails or forks wrong | Autonomous entries skipped/rejected in all selection paths |
| Crashed agent leaves session busy+GC-exempt forever | Marker clears on `exit` as well as `terminated` |
| Stop reports success, cron turn keeps running; failed stop would clear state | Stop cancels the engine turn; failures keep state and return `success: false` |
| Engine turn during `loadSession` suppression dropped permanently | Autonomous-identity updates bypass replay suppression |
| Monitor / idle-reset can't see the engine turn | Both read the activity marker |

## Test plan

- `apps/cli`: 148 targeted tests passing (session-fork, edit-and-resend, execution-service incl. new cancel-path cases, machine-runtime engine-turn, live-status predicate, transient store).
- `packages/components`: 15 stream-item tests passing; tsc clean everywhere; prettier clean.
- Re-verified after the divider revert: remaining diff is guard-only.

- **Review focus:** `session-edit-and-resend-service.ts` and `session-chat-interface.tsx` backward-walk skip conditions, `session-fork-service.ts` `cloneHistoryThroughTurn` rejection, `session-execution-service.ts` engine-turn cancel branch (success ordering), `message-handler.ts` `enqueueACPUpdate` suppression ordering, `machine-runtime.ts` exit wiring.
- **Decisions to challenge:** clearing the marker immediately after a *delivered* cancel (self-heals via turnEnded/re-mark); treating any stale stop with an active engine turn as targeting it; replay-bypass keyed on autonomous identity as the replay-vs-live discriminator.
- **Plausible failures / evidence gaps:** the replay-window reorder (3 lines) has no dedicated harness test — MessageHandler construction is disproportionate; both predicates it orders are independently covered. Dispatching a user message during an engine turn still relies on engine-side queue/reject semantics (pre-existing).

### Authoring context

- **User goal / directives:** land the lifecycle guards as a companion to #653; revert the divider UI after the index-contract finding and track the UI design in a follow-up issue.
- **Constraints / non-goals:** no stream-index contract changes; no replay/import classification rework; no cron-prompt surfacing.
- **Risk-bearing decisions:** engine-turn cancels go through plain ACP `session/cancel` (cancels the agent's current turn, whatever it is); replay bypass trusts the autonomous marker as proof of liveness.
- **Destructive or irreversible behavior:** none — all paths fail closed (marker kept on failure; fork rejected).
- **Deliberately not done or tested:** origin-label UI (needs a contract-safe design); end-to-end Stop click in a live app (covered at service level).
- **Unknowns / confidence:** high confidence from unit coverage of every branch; the two no-harness spots are noted above and bounded to thin wiring over tested predicates.

### Original user prompt

<details>

````text
对，所以你要不要看一下「新条目会进入多少既有生命周期判定」这方面的东西，看看你还有没有遗漏
````

</details>


tommy0103:
Verified and fixed: `Session.emit('terminated')` fires only on the explicit `terminate()` path, while unexpected agent death reports solely through `exit` — exactly the marker-leak case described. `machine-runtime` now also clears the engine-turn marker on the manager's `exit` event (commit ea4b53ae), covered by `tests/machine-runtime-engine-turn.test.ts` which asserts the exit lifecycle path clears the marker.

tommy0103:
Valid — the fork busy predicate (`isSourceBusy`) resolved through `resolveSessionLiveStatus`, which didn't know about the engine-turn marker, so a fork during an engine turn would take the idle-source path and skip `supportsActiveTurnFork` / resident-ACP checks in `forkInner`. The marker is now an input to the shared predicate itself (commit 420280be): the live-status RPC, `isSourceBusy`, and any future caller read it from one place, with presence still outranking. Covered by two new predicate tests (engine turn → running; presence outranks).

tommy0103:
Confirmed against the cancel path: the unfinished autonomous entry flows through `resolveActiveAssistantTurnId` into the Stop control, while `cancelSession` exact-matches only client-turn registries — the request hit the stale branch, returned `success: true`, and the engine turn kept running. Fixed in 885f3715: when a stop request matches no client turn and an engine-opened turn is active, it is cancelled through the plain ACP `session/cancel` (the only live work on the agent) and the activity marker clears; if the agent is already gone, the exit path from ea4b53ae covers cleanup. Two new tests: stop routes to the ACP cancel + marker clears; no engine turn → stale no-op preserved.

tommy0103:
Both fixed in b6471c66:

1. **Failed engine-turn stop:** the cancel path now clears the activity marker only after a delivered ACP cancel. A failed cancel (or an unavailable ACP session) keeps the marker and returns `success: false` — status and the GC guard keep seeing the work, and "cannot send" is distinguished from "cancelled". Two new tests cover cancel-throws and agent-unavailable.
2. **Replay window:** `enqueueACPUpdate` now resolves the autonomous identity *before* `recordSuppressedAcpReplay`, so a live engine turn firing during the `loadSession` suppression window is never dropped as replay — replay notifications carry no autonomous marker, only live engine turns do. A turnEnded falling in the window likewise lands, so the turn can't go permanently missing.

tommy0103:
Fenced in 19e71766: the check-cancel-clear sequence now runs inside the session rewrite barrier (the same primitive dispatch/steer respects), and the activity marker is rechecked inside it. A client turn can no longer start between the ownership snapshot and the awaited `session/cancel`, so the cancel can't land on a fresh client prompt, and presence can't be cleared out from under it; an engine turn that ended in the gap is not cancelled twice. New test: stale-snapshot recheck skips the cancel entirely.

tommy0103:
Fixed in 35a8bc70: `getExecutionSnapshot` now includes the engine-turn activity marker in `hasActiveTurn` (with an explicit comment that this is live work on the same agent). `SessionDispatchWatcher`'s admission sees the session as active (`active-session` noop) while an engine turn runs, so a queued user message stays queued; when the turnEnded marker clears the activity, the next check chain dispatches it. Two snapshot tests added (engine turn → `hasActiveTurn: true` with no `activeTurnId`; cleared → false). This also removes the reliance on engine-side `another turn is already in progress` rejections as the only backstop.

Leeeon233:
Hi @tommy0103, could you resolve the conflicts?

PR #668, opened 2026-09-13
fix(cli): settle acknowledged steers when their target turn stops

### LINKED ISSUE #666: [Bug] Timed-out steer remains pending_apply after its target turn is stopped
### Affected area

Agent runtime / ACP

### Installation method

Built from source

### Lody version or commit

The exact revision used for the live observation was not recorded. The affected control flow is present at `b5746d02755257331a6856b4142f02fe1d747ab1`.

### Operating system

macOS 26.5.1 arm64

### Agent or runtime

Built-in Codex runtime with the acknowledged-steer capability. Provider session and model identifiers are intentionally omitted.

### What happened?

When a user submits guidance while a turn is active, Lody writes the user entry as `pending_apply` and offers it through `session/steer`. If the local steer RPC times out before the ACP adapter reports either application or a definitive refusal, the UI keeps the entry in `pending_apply` and continues to show **Steering**.

Stopping the target turn does not resolve that entry. The active prompt is cancelled and the session returns to idle, but the steer is not marked applied, failed, or cancelled and is not promoted to ordinary dispatch. Because ordinary dispatch deliberately skips `pending_apply`, the user's intent is stranded indefinitely.

A separately submitted normal message can still run afterward. It is a new user turn; it does not replay or recover the stranded steer.

The resulting entry is an orphaned state:

- it is absent from the provider conversation;
- it is absent from the ordinary dispatch queue;
- it has no terminal failure or cancellation state;
- it remains displayed as **Steering** after its target turn has ended.

### What did you expect?

Every steer associated with a stopped target turn should converge to an observable, recoverable state without risking duplicate delivery:

- a confirmed application should transition to the successor turn;
- a confirmed refusal or pre-delivery cancellation should return the same user turn to ordinary dispatch;
- an unknown delivery result should be represented explicitly and offer a safe reconciliation or retry path instead of remaining **Steering** forever.

Stopping a turn must also settle or transfer ownership of its outstanding steer requests and application waiters. A late result must not resurrect stopped state or overwrite a newer user turn.

### How can we reproduce it?

Use synthetic prompt text; no original conversation data is required.

1. Start an active turn through `SessionExecutionService` using an ACP client that advertises acknowledged steer.
2. Submit a steer such as `Use the alternate implementation.`
3. Hold the steering extension request unresolved and do not emit the steer-application notification until after the UI's five-second RPC timeout.
4. Confirm that the user entry remains `pending_apply` and the UI shows **Steering**.
5. Stop the target turn and wait for the session to become idle.
6. Observe that the steer remains `pending_apply` and is not selected by ordinary dispatch.
7. Submit a separate normal message. It can execute, while the original steer remains stranded.

The live behavior was observed once. The state transition should be reproducible deterministically with a controlled unresolved ACP steer request.

### How often does it happen?

Only once in a live source build. The affected path is deterministic when the ACP steering request remains unresolved across cancellation.

### Additional context

The likely failure chain spans these existing ownership boundaries:

1. `workspace-machine-rpc-facade.ts` applies a five-second timeout to `requestSessionSteer`.
2. `use-session-actions.ts` promotes only the definitive `no-active-turn` disposition; timeout and ambiguous errors intentionally leave the entry in `pending_apply` to avoid duplicate delivery.
3. `AgentClient.steerPrompt` waits for both the steer submission and original prompt completion before rejecting an unapplied waiter. Cancelling the prompt is insufficient when the steer submission never settles.
4. `pendingPromptCompletion` tracks ordinary prompt requests but not steer submissions or `steerApplicationWaiters`, so Stop can finish without settling the steer.
5. Session dispatch skips `pending_apply`, leaving no recovery path after the target turn ends.

The no-blind-resend rule is correct: the [OpenAI Responses steering contract](https://developers.openai.com/api/reference/cli/resources/beta/subresources/responses) distinguishes acceptance from application, uses the successor `response.created` event as the commit point, and treats a missing acknowledgement as an unknown outcome. The bug is that Lody does not reconcile or expose that unknown outcome when the target turn is stopped.

Related but distinct:

- #477 covers pending steer waits that can block a later user message. In this report, a later ordinary message remains dispatchable; the original steer itself is orphaned.
- #530 requests broader state feedback and recovery after interruption. This report identifies the concrete `pending_apply` lifecycle gap.

No screenshot, raw log, prompt, conversation excerpt, session identifier, user identifier, workspace identifier, repository content, or sensitive path is included in this report.

### Before submitting

yetuge:
## Related issue

Closes #666

## Problem / pressure

A steer submitted while a turn is running must not outlive its target turn, but `steerSessionLocked` waited for the agent's application verdict with no bound tied to that turn:

- Request-transport runtimes settle the verdict only when the agent answers the held extension request or the connection closes — stopping the turn aborts neither — so after a Stop the await hung forever: the steer response never resolved, the guide stayed in `pending_apply` (a status ordinary dispatch deliberately skips), and the per-session steer mutation queue stayed blocked behind the hung await.
- Prompt-transport runtimes reject the verdict directly when the stopped prompt rejects, which fell into the generic `error` disposition that also left the history entry untouched.

Either way the user's message was stranded exactly as described in #666: absent from the provider conversation, absent from ordinary dispatch, with no terminal state.

## Summary

`steerSessionLocked` now races `steerRun.applied` against the target turn's prompt-run settlement (the prompt run is the cancellation owner, so its settlement is the structural bound — no arbitrary timeout):

- When the turn settles first, the guide is settled to a terminal history status via the existing `setTerminalUserTurnStatus` (`canceled` when a stop was requested, `failed` when the turn ended without an answer) and the response is `stale-turn`. The catch path applies the same settlement when a post-submission stop rejects the verdict, so both transports converge.
- Confirmed refusals (`AgentSteerNotDeliveredError`) and pre-submission rejections keep the existing requeue-to-dispatch behavior unchanged; the terminal path only covers the unknown post-submission window.
- Unknown delivery results are deliberately **terminal rather than re-queued**: the provider may still take the held request, and replaying could duplicate it. The application waiter stays registered in the agent client until the agent answers or the connection closes, so a late verdict cannot resurrect the stopped turn.
- A bilingual Agent Note (`implemented/bug-fix/2026-09-13-steer-settles-with-stopped-turn` + `.zh.md`) records the decision per `AGENTS.md`.

## Visual explanation

```mermaid
flowchart TD
    A["steerSessionLocked<br/>submits steerPrompt"] --> B{"Promise.race"}
    B -- "verdict resolves" --> C["existing application path<br/>(ownership handoff, unchanged)"]
    B -- "verdict rejects:<br/>AgentSteerNotDeliveredError" --> D["requeue to ordinary dispatch<br/>(confirmed refusal, unchanged)"]
    B -- "prompt run settles first<br/>(stop or end)" --> E["terminal settlement:<br/>canceled if stop requested, else failed;<br/>response stale-turn"]
    B -- "verdict rejects otherwise<br/>+ stop requested" --> E
    B -- "verdict rejects otherwise,<br/>no stop" --> F["error disposition<br/>(unchanged)"]
```

Simple-change justification: the diff is 347 changed lines but they are one control-flow change in a single function plus its regressions; the Mermaid view above is the complete decision surface.

## Before / after

| Before | After |
| ------ | ----- |
| Stop with a held steer request: `steerSession` never returns, guide stays `pending_apply`, steer mutation queue blocked | Response resolves `stale-turn`, guide settled `canceled`, queue unblocked |
| Prompt-transport stop rejects the verdict: `error` disposition, entry untouched | Same terminal settlement as above |
| Confirmed refusal / pre-submission rejection | Unchanged (requeue to dispatch) |
| Late application after stop (verdict arrives later) | Waiter resolves unconsumed; no resurrection |

## Test plan

- `npx vitest run tests/session-execution-service.test.ts` (apps/cli): 111 passed, including two new regressions that reproduce #666 — a held verdict plus a stopped turn now resolves `stale-turn` and writes the `canceled` status (previously the test timed out at 30s with the response pending), and a rejected verdict on a stopped turn converges to the same settlement (previously `error` with no history write). Both new tests fail on the parent commit.
- Full `apps/cli` suite with and without the change: identical failure set (88 FAIL lines, name-by-name; 61 failing tests are pre-existing on this Windows machine — real-process login flows, native binary downloads, `better-sqlite3`-adjacent suites), 0 new / 0 disappeared; the with-change run additionally passes the two new tests.
- `oxlint` (repo gate is `--quiet`): 0 errors on both changed files; `prettier --check` clean; `pnpm run docs status` / `docs check` output for the two new notes shows `status: implemented`, `translation: current`, `missingTranslation: false`. `pnpm run typecheck` in apps/cli reports the same three pre-existing cross-package module-resolution errors as the parent commit (untouched packages: `@lody/shared/platform-kind` subpath, `better-sqlite3` types).

- **Review focus:** `steerSessionLocked` in `apps/cli/src/session/session-execution-service.ts` — the `Promise.race` bound and the two settlement branches (post-race and catch); confirm the requeue path for `AgentSteerNotDeliveredError` and the pre-submission rejections are untouched.
- **Decisions to challenge:** terminal-instead-of-requeue for unknown post-submission results (duplicate-delivery risk vs. visibility); marking a non-stop settlement `failed`; leaving the agent-client waiter registered until answer/connection close.
- **Plausible failures / evidence gaps:** verified against the service contract and mocked agent clients only — no live Codex/Claude adapter run; the `failed` status for a non-stop settlement is a judgment call, not a documented product decision.

### Authoring context

- **User goal / directives:** automated contribution workflow: pick an unclaimed agent-core issue, verify the root cause at current HEAD, fix minimally with regression tests, and open a focused PR.
- **Constraints / non-goals:** no behavior change for confirmed refusals, pre-submission rejections, or the application path; no timeout introduced; no agent-client protocol changes.
- **Risk-bearing decisions:** terminal statuses replace an infinite wait; if a provider later applies a steer after its turn was stopped, the history shows `canceled` while the agent may execute — matching the report's requirement that a late result must not resurrect stopped state.
- **Destructive or irreversible behavior:** none; history entries only move from `pending_apply` to a terminal status.
- **Deliberately not done or tested:** live adapter verification (no provider credentials on the authoring machine); no changes to `agentClient.steerPrompt` abort wiring.
- **Unknowns / confidence:** high confidence in the response bound and settlement; medium on the `failed`-vs-`canceled` distinction for non-stop settlements.

### Original user prompt

<details>

````text
按 G:\agent学习资料\pr\定时任务提示词.md 完整执行一轮
````

(Verbatim triggering instruction. It references a local scheduled-task prompt file that directs an
automated run: follow up on tracked pull requests, then — when quality gates pass — claim one
unclaimed issue in an agent-core TypeScript repository, fix it with regression tests, and open a
pull request. This PR is that run's output for issue #666.)

</details>


yetuge:
Both findings confirmed as real — thank you for the precise analysis. Fixed in 3aa4fbc.

**1. Preserve provable refusals after the settlement race**

The race branch now attaches a non-blocking consumer to the pending verdict right after the terminal write (the response itself still never waits on the held request, so the original stranding stays fixed). When the verdict later rejects with `AgentSteerNotDeliveredError` — which the agent client only produces from the agent's own invalid-request answer, so it is a proof of non-delivery — a new `requeueSteerAfterLateRefusal` returns the guide to ordinary dispatch: it flips back the terminal status that *this same race branch* wrote (no other path writes `canceled`/`failed` over a steer entry whose verdict is still pending), then re-walks the ordinary requeue guards (not the active turn, not already handled) and rewrites the `latestUserMsgId` pointer. Ambiguous rejections (closed connection, dead process) and late acceptances keep the terminal status, so nothing that may have been taken is ever replayed.

**2. Release the lease from a late steer acknowledgement**

The same consumer releases a late acceptance's lease immediately: `steerRun.applied.then((lease) => lease.release())`. `handleSteerApplied` installs `waiter.released` as `steerApplicationBarrier` before the consumer runs, so the window in which later `sessionUpdate`s block is bounded to microtask ordering — the barrier is cleared as soon as the release resolves, and ownership is not transferred (the guide stays terminal; no replay of a taken steer).

**Regression tests** (new, in `tests/session-execution-service.test.ts`):

- `requeues a settled steer whose late verdict proves the agent refused it` — fails on 855eced (entry stays `canceled`, no pointer write), passes now (entry back to `pending` + `latestUserMsgId` rewritten).
- `releases the lease of a steer acknowledgement arriving after settlement` — fails on 855eced (`release` never called → the barrier would wedge every later `sessionUpdate`), passes now.

Full `apps/cli` suite: 2477 tests, 61 failures — all identical to the pre-existing Windows environment failures recorded at the parent commit (list in CI logs matches by name; zero new failures). oxlint 0 errors, prettier clean, docs check reports no new findings (the pre-existing `apps/cli/src/session/AGENTS.md` size-gate warning is unchanged), and both language versions of the implementation note were updated to replace the "late verdict resolves into a promise nobody consumes" assumption this review correctly refuted.

yetuge:
Both findings verified and fixed in 78f90ed:

**1. Requeue before history synchronization** — correct. When the steer's history entry has not synced to this daemon, `queueable` stayed false and the early return left the race branch's terminal-without-entry record in place without publishing the pointer, so the late-syncing entry was repaired back to terminal and the guide stranded. The requeue now covers the entry-absent case: it requeues through the pointer exactly like the ordinary refusal path treats a missing entry, and drops the stale record so the late-syncing entry dispatches via the `pending_apply` pointer match in `findNextDispatchableUserTurn` instead of being repaired to terminal.

**2. Preserve newer activations** — correct, and it is the same rule `setDispatchProcessing` already documents for execution-side writes (execution claims only its own `processingUserMsgId` slot). The pointer publish now treats the slot as occupied while `latestUserMsgId` names a live activation — anything that is not this steer itself and is not retired by `lastHandledUserMsgId`, `lastMissingHistoryUserMsgId`, or `settledActivationUserMsgId` — and leaves it alone. The flipped `pending` entry is then picked up by the chronological scan that the newer activation keeps running, and a not-yet-synced entry keeps steer intent instead of destroying the newer turn's only activation signal.

Tests: five new cases — entry absent → pointer requeue + stale record cleared; entry absent while a newer activation holds the pointer → no pointer write, record still cleared; entry present under a live newer activation → flip to `pending` without a pointer rewrite; pointer naming an already-handled turn (consumed activation) → rewrite still happens; pointer retired by `settledActivationUserMsgId` → rewrite still happens. The first three fail on the previous head.

yetuge:
All three findings verified and fixed in 7fc09af:

**1. Retain activation for an unsynced refused steer** — true. With the pointer owned by a live activation and the entry still absent, dropping the record left nothing durable, and a late-syncing `pending_apply` entry dispatches only through a pointer match, so the proven-undelivered guide stranded. That corner now keeps the terminal-without-entry record: the late entry is repaired to its terminal status — visible non-delivery, the same terminal settlement the race branch already reported — instead of stranding as steer intent nothing dispatches. Auto-activating it instead would require a queued-activation slot in session meta (schema + `hasPendingUserTurnActivation` + watcher-unload coordination); we deliberately did not introduce that unilaterally and are happy to follow up if you consider it warranted.

**2. Recheck the dispatch pointer after awaited history work** — true, and the guard was worse than merely late: computed before two awaits, it was exactly the read-await-rewrite the session contract forbids. The pointer decision now runs on a meta re-read after `updateHistory` settles (with a fresh `lastHandledUserMsgId === userTurnId` re-check), so publications during the awaited history work are observed. The residual window between that re-read and the `upsertSessionMeta` patch is the same one `requeueUndeliveredSteer` — the ordinary refusal path — already operates within; eliminating it entirely needs a compare-and-swap in the meta repo layer, which we did not unilaterally add.

**3. Preserve unrelated missing-history tombstones** — true. The patch now includes `lastMissingHistoryUserMsgId: undefined` only when it names the requeued steer itself; otherwise the key is omitted and the older turn's permanent one-shot ack survives (the meta store deletes on an explicit `undefined` and no-ops on an omitted key).

Regression tests (the first three fail on the previous head, all four pass now): absent entry under a live activation keeps the record; an activation published inside the awaited history write is observed by the pointer decision; a tombstone naming an older turn survives the pointer write; a tombstone naming the requeued steer itself is still cleared. The file's suite is 121/121; the full `apps/cli` run shows no new failures beyond the known Windows-environment flake family (the `CodeCollabV2Service` temp-dir `EBUSY`/`better-sqlite3` set drifts run-to-run and is untouched by this commit); `tsgo`/`oxlint`/prettier are clean on the touched files. The bilingual agent note now describes the revised requeue semantics.

One scope note, consistent with the previous reply: `requeueUndeliveredSteer` shares both the unconditional tombstone clear and the read-await-rewrite shape. As before, we kept it untouched as pre-existing ordinary-refusal semantics and leave promoting the fix there to you.

yetuge:
Heads-up after pushing 7fc09af: the reconcile-policy check now reports the PR at 1146 changed lines, which per `CONTRIBUTING.md` needs **a maintainer assignment on the linked issue (#666) before review**. Could a maintainer assign #666 (or grant an exception) so this can proceed through review? The three review rounds so far are all addressed in the latest commits.

yetuge:
Both findings verified against `7fc09af`. I have deliberately not pushed a fix yet, and I would rather explain why than leave them unaddressed.

**1. Persist terminal settlement before returning stale-turn** — confirmed real. The race branch's `canceled`/`failed` write lands in `terminalTurnStatusWithoutEntry`, an in-memory per-daemon map. If the daemon restarts before the entry syncs while `latestUserMsgId` still names the steer, the late `pending_apply` entry matches the pointer, so `findNextDispatchableUserTurn` dispatches it again — a second submission of a guide the provider may already have taken before the stop. That is a durability gap, not a style point.

**2. Requeue the refused steer behind the newer activation** — also confirmed, and it is the mirror image of the previous round's finding #1. Dropping the record stranded the guide (previous round); keeping it leaves the guide terminal forever whenever another activation owns the pointer (this round). Both behaviours the single `latestUserMsgId` slot can express are now shipped, and each leaves a gap — because a `pending_apply` entry is dispatchable only through a pointer match.

Closing either finding therefore needs a second, durable channel: a queued-activation slot for a refused steer, or a durable terminal marker for the race branch. Either way it is a new field in the shared `SessionMeta` (`packages/shared/src/schema.ts`), its `getPendingUserTurnActivationId` / `hasPendingUserTurnActivation` predicate — documented as the ONLY pending-turn predicate — and the corresponding watcher-unload coordination. That is a contract change, not a bug fix contained in `session-execution-service.ts`, so I would rather the maintainers choose the direction than pick one unilaterally.

There is also a practical constraint: the PR is at 1146 lines and `reconcile-policy` requires a maintainer assignment on #666 before review, so every additional commit deepens that violation.

If a maintainer confirms the direction, I can prepare the durable queued-activation slot (schema + predicate + watcher lifecycle + regression tests) on top of this branch. Alternatively, if you would rather keep the slot ownership exactly as documented in `apps/cli/src/session/AGENTS.md`, I can persist the race branch's terminal marker instead and leave the pointer contract untouched. Happy to follow whichever you prefer.

PR #693, opened 2026-09-14
fix: preserve pending input across turn interrupt

wibus-wee:
## Problem

Conversation Stop / ESC must interrupt only the exact active turn. When another input is waiting as a steer, interruption alone does not prove whether the provider consumed it: replay can duplicate input, while dropping proven non-delivery can lose it.

## Scope

- Keep the adapter's three delivery outcomes: `applied` is never replayed; `not-applied` can become ordinary pending dispatch; `unknown` stays `pending_apply` without automatic replay.
- Separate pending-input policy from pre-prompt ACP process lifetime. Binding a resident session no longer unconditionally marks its process for termination.

| Cancellation | Pending input | Pre-prompt ACP session |
| --- | --- | --- |
| User Stop / ESC | promote | discard |
| Edit & Resend | preserve | keep |
| Access revocation | preserve | discard |

- Preserve create/restore termination fences during initialization, then release them so they cannot override a later `keep` during config. Retain foreground configuration's owner `AbortSignal`.
- Keep Edit & Resend's prepare → cancel → wait → commit → adopt sequence safe for generic ACP processes; do not assume a prepared session survives process death.
- Report failed durable promotion as `promotion-failed`, distinct from uncertain provider delivery. If CLI history promotion succeeded but `latestUserMsgId` failed, the client repairs the pointer and dispatch for the same `pending` / `seen` input. Running, terminal, removed, and delivery-unknown inputs are not replayed.

Adapter [#43](https://github.com/LodyAI/acp-extension-codex/pull/43) has merged. This PR now pins follow-up [acp-extension-codex#44](https://github.com/LodyAI/acp-extension-codex/pull/44) at `38f29ed`; merge that adapter PR before this PR. Durable Session Hold, SQLite control state, Goal/Operation hold, durable Resume, and a new mutation lane remain out of scope.

### Codex ambiguous-response reconciliation

The inner adapter catch still inferred non-delivery from an ended turn and removed steer identity before notification drain. The follow-up retains the submitted identity through prompt completion and performs one bounded history read after a failed ambiguous response:

```text
failed turn/steer response -> drain received notifications
  matching user item -> one applied acknowledgement -> injected
  explicit refusal  -> failed (replay-safe)
  otherwise         -> thread/read(includeTurns: true), original thread/turn
                         matching clientId -> applied -> injected
                         absent/error/timeout -> reject (unknown)
```

Drain/read share five seconds; live evidence can finish while history is pending. A late read has no acknowledgement side effect. Stop and terminal turns are not negative proof. There is no resend, new turn, or resume. Successful ACKs keep the existing live-notification path; this does not eliminate all unknown outcomes or implement cross-restart/idempotent recovery.

## Behavioral coverage

- ESC during config waits for real `Session.terminate()` cleanup; B uses a new resident client.
- Edit & Resend during config runs the real edit and execution services and adopts the replacement in its original live process, for both resident and just-restored sessions.
- Access revocation discards the pre-prompt process without promoting pending input.
- Injected metadata failure after history promotion returns `promotion-failed`; already-pending recovery repairs the activation pointer and B runs once.
- Late `applied` after Stop never replays B; `unknown` never auto-promotes.

Lifecycle tests control only the ACP/OS boundary, with process-owned session identities invalidated on exit. They do not replace `Session.terminate()` or `cancelSession()` with no-ops. Ordering uses explicit deferred signals, not sleeps. These are deterministic service-level tests, not live-provider E2E.

Negative controls were also run: restoring the old unconditional termination and swallowed promotion errors makes the two blocker regressions fail; disabling pending/seen client repair makes all four corresponding recovery cases fail. The fixes were restored before final validation.

### Current follow-up: `1e5109e9`, adapter `38f29ed`

- Adapter full suite: 630 passed, 27 existing gated skips, no retries.
- Restoring the old catch and prompt cleanup makes three selected regressions fail: persisted application, queued notification, and unknown after completion. With the fix restored, all 21 steer tests, typecheck, build, and diff checks passed again.
- A real Codex 0.153.4 prompt smoke returned `end_turn` on a separate rerun; the first launch timed out. This smoke does not inject a lost steering response.
- Main workspace typecheck, lint, i18n, and code-collab/platform/public boundary checks passed. The first full local `pnpm check` was stopped after discovering inherited `NODE_ENV=production` makes React `act` undefined; a direct probe confirms it is a function under `NODE_ENV=test`. The unchanged representative UI suite passed all four tests under the correct environment. A full `test:ci` rerun with `NODE_ENV=test` and an isolated CommonJS `TMPDIR` passed shared (1,227 tests) and progressed through UI tests without failures, then was stopped once the exact-head formal full CI passed. Neither interrupted local full command is claimed as green.
- Formatting was run; its unrelated Electron-only formatting change was reverted. Documentation and diff checks passed.
- GitHub exact HEAD `1e5109e937ebfec2b6e18eaa30b7d65eb5418e83`: [Static checks and Tests passed](https://github.com/LodyAI/Lody/actions/runs/34990367646); [Desktop E2E smoke passed](https://github.com/LodyAI/Lody/actions/runs/34990367660). All contribution-policy/scope checks also passed. These are current-head results, not reused previous-head results.

### Prior lifecycle/promotion fix: `7c1f3e20`

- Fix commit: `7c1f3e20`.
- Targeted CLI execution / edit / dispatch suites: 174 tests passed, including both resident and restored lifecycle cases.
- Component session actions: 46 tests passed; shared message schema: 53 tests passed.
- Workspace typechecks, lint, i18n checks, and code-collab/platform/public boundary checks passed.
- Formatting, documentation checks, and `git diff --check` passed.
- Full local `pnpm test:ci` passed the other workspace suites (shared: 1,227 tests; components: 3,697 tests) and 264 CLI files, but stopped at five `gh-shim-script` failures. Diagnosis: the host's `/tmp/package.json` has `type: module`, so Node interpreted generated CommonJS shim fixtures as ESM (`require is not defined`). With an isolated CommonJS `TMPDIR`, that unchanged suite passed all 7 tests. Electron's interrupted follow-on suite was run separately: 112 tests passed. The full command itself is not claimed as a green run.
- All 174 affected CLI tests were rerun after the final restored-session fence adjustment; typecheck and type-aware lint were also rerun for that adjustment.
- GitHub at the prior HEAD `7c1f3e20`: Static checks, Tests, and Desktop E2E smoke passed. These results apply only to that prior HEAD.


wibus-wee:
@copilot resolve the merge conflicts in this pull request

PR #699, opened 2026-09-14
test: align failure reporter coverage with PR-only use

wibus-wee:
## Problem / pressure

After Daily reporting moved all evidence to Actions artifacts, the attachment report generator remained shared in name and defaulted to an unreachable Daily mode. Most of its tests therefore exercised behavior that no workflow called.

## Summary

- Rename the attachment reporter and its tests for their sole remaining PR workflow consumer.
- Remove the dead Daily channel branch and emit PR markers and subjects directly.
- Update fixtures and attachment ownership assertions to use PR paths and markers.
- Preserve the existing PR manifest schema, attachment validation, and failure handling.

## Visual explanation

Simple change: one existing helper is narrowed to its only caller; runtime data flow is unchanged.

## Before / after

| Before | After |
| ------ | ----- |
| Reporter defaulted to an unused Daily mode | Reporter is explicitly PR-only |
| Most fixtures and markers described Daily attachments | Fixtures and markers match the PR workflow |
| PR workflow selected the shared channel at runtime | PR workflow invokes the PR-specific reporter directly |

## Test plan

- No tests were run at the user request; hosted CI is the verification boundary.
- Targeted Prettier formatting completed for all changed JavaScript and Markdown files.
- The repository-wide format command could not start because `corepack` is absent from the current PATH.
- The documentation check did not return within the local command window.


wibus-wee:
Closing because this follow-up scoped the attachment reporter to PR failures, while the requested behavior and coverage are specifically for Daily artifact-only reporting. PR #698 already owns that Daily change; this PR is therefore out of scope.

PR #761, opened 2026-09-16
fix: prevent public browser navigation feedback loop

wibus-wee:
## Problem / pressure

The desktop public browser can feed native URL observations back into the renderer as fresh navigation requests. Redirects such as baidu.com then cause repeated loadURL calls and apparent refresh loops; the superseded load reports ERR_ABORTED (-3), not connection reset.

## Summary

- Separate explicit address-bar navigation requests from native URL/state observations.
- Keep redirects, history, same-document changes, and page-opened destinations out of the explicit navigate path.
- Memoize the Electron bridge so state updates do not rebuild view effects.
- Add a controller regression test and record the diagnostic and fact-check evidence.

## Visual explanation

Simple change: the key behavior is a one-way ownership boundary; the existing component layout would not make the fix clearer than the request/state flow described above.

## Before / after

| Before | After |
| ------ | ----- |
| Native URL state could be treated as a new navigation intent. | Only an explicit `{ id, url }` request can call `navigate`; native state updates display/history. |
| Renderer state changes could recreate bridge-dependent effects. | The Electron bridge identity is stable for the surface lifetime. |

## Test plan

- `pnpm --filter @lody/components typecheck` — passed.
- `pnpm e2e:build` — passed.
- Xvfb-backed `pnpm e2e:smoke` — 5/5 P0 scenarios passed.
- Built Electron diagnostic for baidu.com — one main-frame navigation after the redirect, no second navigate or loop cancellation.
- `pnpm run docs check` and formatting checks — passed; repository reported only pre-existing warnings.
- Focused/full component Vitest remains blocked by the existing React 19 environment error: `act is not a function`.

- **Review focus:** Review the request/state ownership in `public-browser-surface.tsx` and `session-browser-panel.tsx`, plus the controller regression test.
- **Decisions to challenge:** Confirm that native redirects and history movements should update the visible address without becoming new renderer navigation intents.
- **Plausible failures / evidence gaps:** The active E2E registry has no browser-navigation journey, and component Vitest is blocked by the existing React 19 `act` export mismatch.

### Authoring context

- **User goal / directives:** Stop the right-side public browser from repeatedly refreshing when navigating external URLs.
- **Constraints / non-goals:** Keep native browser ownership and page-opened security behavior unchanged; do not add live-network E2E fixtures.
- **Risk-bearing decisions:** Explicit navigation uses a monotonic request identity; native state remains authoritative for the displayed URL and history.
- **Destructive or irreversible behavior:** None; the change only changes navigation synchronization and adds tests/notes.
- **Deliberately not done or tested:** No live-network journey was added because the repository E2E policy excludes it; full component tests cannot run in the current environment.
- **Unknowns / confidence:** High confidence in the feedback-loop fix from the real built Electron diagnostic; loading-state churn on direct `www.baidu.com` remains separately observable.


packages/components/src/components/sessions/session-browser-panel.tsx
```diff
@@ -1040,8 +1048,7 @@ function SessionBrowserPanelController({
       ) : currentAddress?.engine === 'public-web' ? (
         <PublicBrowserSurface
           browserId={`session-browser-${session.id}`}
-          url={currentAddress.logicalUrl}
-          navigationRequestId={publicNavigationRequestId}
+          navigationRequest={publicNavigationRequest}
```
wibus-wee, line 1063: 🔴 **Stale navigation replays after remount**

When `PublicBrowserSurface` remounts, `publicNavigationRequest` replays the last explicit URL despite no new request. Cancelling remote-preview confirmation remounts the surface, resetting redirects or native history to that stale URL.

<details>
<summary>Learn more</summary>

The request remains non-null after `PublicBrowserSurface` dispatches it. The surface normally suppresses duplicates with its local `navigationRef`, but that ref resets whenever the component unmounts. Starting managed navigation replaces the public surface with the progress view; a remote confirmation or resolution failure then renders a fresh public surface with the old request. That fresh instance calls `navigate` with the retained URL even though the user created no new public navigation intent.

**Example:** A user enters `example.com`, which redirects to `www.example.com/account`. They then enter a remote loopback preview address and cancel its confirmation. The public surface remounts and navigates back to `https://example.com/`, discarding the redirected page and adding an unwanted navigation.

**Recommended fix:** Add a request-consumed callback from `PublicBrowserSurface` to `SessionBrowserPanelController`. Clear only the matching `publicNavigationRequest` when the surface has accepted it for dispatch. Keep an undispatched request across remounts so a remount before `surfaceReady` does not lose genuine user intent.

</details>


<a href="https://app.devin.ai/review/lodyai/lody/pull/761" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

wibus-wee (commented):
**Devin Review** found 1 potential issue.


<a href="https://app.devin.ai/review/lodyai/lody/pull/761" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

PR #767, opened 2026-09-17
fix: render Devin ACP subagents as task lifecycle entries

### LINKED ISSUE #765: [Bug] Devin ACP subagents render as instantly-completed tool calls with no lifecycle or result
### Affected area

Agent runtime / ACP

### Installation method

Desktop release

### Lody version or commit

0.94.0

### Operating system

macOS (arm64) — affects all platforms

### Agent or runtime

Devin CLI via ACP (`devin acp`), registry binary distribution, version 3000.10.21

### What happened?

In a Devin CLI session, when the agent spawns subagents (`run_subagent`), each one renders as an ordinary tool call that completes instantly with placeholder text — `Ran explore subagent …` / `"Background subagent started."` — and later `read_subagent` polls render as `Checked on subagent …` / `"Subagent completed."`. There is no task lifecycle, no progress, and the subagent's final result never reaches the transcript.

Verified at the protocol level, this is a capability-negotiation gap rather than a rendering bug. Devin gates its structured subagent events behind client capabilities declared at `initialize`:

```json
"clientCapabilities": {
  "_meta": {
    "cognition.ai/subagentSupport": true,
    "cognition.ai/subagentControl": true
  }
}
```

Lody does not advertise them today (Devin's own log confirms `subagents=false, subagent_control=false`), so `devin acp` degrades to flat tool calls:

```mermaid
sequenceDiagram
    participant L as Lody client
    participant D as devin acp
    L->>D: initialize, no subagent capability
    D->>L: tool_call run_subagent, status completed, content Background subagent started
    D->>L: tool_call read_subagent, status completed, content Subagent completed
    Note over L: two instantly completed empty tool calls, no lifecycle, no result
```

With the flags advertised, the same session instead emits standard `session/update` notifications whose `_meta` carries the full lifecycle:

```mermaid
sequenceDiagram
    participant L as Lody client
    participant D as devin acp
    L->>D: initialize, clientCapabilities._meta cognition.ai/subagentSupport = true
    D->>L: tool_call_update, toolCallId = subagent agentId, _meta.subagent_started = title, task, profile, model, isBackground, depth
    D->>L: subagent-produced updates tagged _meta.subagent_context = parentAgentId
    D->>L: tool_call_update, same agentId, _meta.subagent_completed = success, summary
    Note over L: maps to SubagentTaskPayload, renders as a real task card
```

### What did you expect?

Devin subagents surface the way other providers' subagent tasks do — a task entry with lifecycle/status and the final summary — instead of two instantly-"completed" tool calls that look finished while the subagent is still running.

### How can we reproduce it?

1. Start a Devin CLI session in Lody.
2. Ask the agent to delegate, e.g. "spawn a background explore subagent that lists the files in this directory".
3. Observe the transcript: the subagent appears as a tool call that is already `completed` with content `Background subagent started.`, and the final answer never shows up.

### How often does it happen?

Every time

### Relevant log output

```shell
# devin acp stderr at initialize (client capabilities as negotiated):
... subagents=false, ... subagent_control=false ...
```

```jsonc
// persisted history items for a spawned subagent:
{ "type": "tool_call", "toolCallId": "run_subagent_N#…",
  "title": "Ran explore subagent …", "status": "completed",
  "content": [{ "type": "content",
    "content": { "type": "text", "text": "Background subagent started." } }] }
{ "type": "tool_call", "toolCallId": "read_subagent_N#…",
  "title": "Checked on subagent …", "status": "completed",
  "content": [{ "type": "content",
    "content": { "type": "text", "text": "Subagent completed." } }] }
```

### Additional context

Suggested shape, matching the existing provider-normalization pattern:

- `apps/cli/src/agent/agent-client.ts`: advertise the `_meta` flags in `initialize` when `agentConfig.agentType === 'devin'` (same precedent as the Grok `clientIdentifier` `_meta` injection).
- New `packages/shared/src/acp/devin-subagent-task.ts` next to `claude-subagent-task.ts`: zod-parse `subagent_started` / `subagent_completed` into `SubagentTaskPayload` (`taskId` = agentId, `actor` = profile, `description` = task, `modelId` = model, `isBackgrounded` = isBackground, `summary`, status mapping), plugged into the existing parser chain in `history-apply.ts`.
- Updates tagged `subagent_context.parentAgentId` must at minimum not leak into the flat transcript as parent-authored messages; nesting the subagent's live transcript under its task card would need a shared-model change and is better scoped as a follow-up that benefits every provider.
- `subagentControl` additionally unlocks `_cognition.ai/subagent/*` requests (list / cancel / foreground); only needed if a control UI is wanted.

Caveat: `cognition.ai/*` is Cognition's private ACP extension surface and is undocumented, so it may drift across Devin CLI versions. Keeping the parser fail-safe (unparseable `_meta` → today's degraded rendering) makes the feature self-healing.

### Before submitting

Astro-Han:
## Related issue

Closes #765

## Problem / pressure

On Devin ACP sessions (`devin acp` from the official Devin CLI binary), spawned subagents render as instantly-completed `run_subagent`/`read_subagent` tool calls with no progress, no lifecycle, and no result text. Devin gates its subagent protocol behind a private `cognition.ai/subagentSupport` client capability that Lody never advertises, so Devin falls back to degraded flat tool calls and the `SubagentTaskPayload` task panel has nothing to consume.

Once the capability is on, the work is not just parsing two markers: Devin tags every subagent-produced update — text, tool calls, usage — with `subagent_context`, and those internals reach several consumers outside the transcript funnel (usage metering, runtime config, plan snapshots, attachments). The history layer also compacts away the `in_progress` lifecycle row before any parser can see it.

## Summary

- `apps/cli/src/agent/agent-client.ts`: for `agentType === 'devin'`, advertise `_meta["cognition.ai/subagentSupport"]: true` at `initialize`. `sessionUpdate` records task ids from lifecycle markers and drops `subagent_context`-tagged **non-tool** updates whose owner is a known task — the single choke point in front of usage/config/plan/title/attachment consumers.
- New `packages/shared/src/acp/devin-subagent-task.ts`: zod-guarded readers for the private markers — `subagent_started`/`subagent_completed` → `SubagentTaskPayload` (agentId→taskId, profile→subagentType, model→modelId, isBackground→isBackgrounded, summary/error), `subagent_context` → owning subagent id (`"root"` excluded), plus a passthrough predicate for unrecognized `subagent_*` payloads.
- `apps/cli/src/lib/acp/history.ts`: `filterNotificationsForHistory` exempts rows that parse as Devin lifecycle markers, the same treatment `_meta.lody.task` snapshots already get — otherwise the started row never reaches the applier.
- `packages/shared/src/acp/history-apply.ts`: the Devin parser joins the lifecycle chain; a context-tagged lifecycle row sets `parentTaskId` for nesting. The applier drops context-tagged updates once the owning task row exists, except tool updates merging into an already-persisted row — permission requests write a pending `tool_call` item before the tool runs, and dropping its terminal update would leave it pending forever. The check lives in the apply loop, not `buildMessageContentFromNotification`, because terminal-output compaction reuses that builder.
- Fail-open under drift: unknown owner ids and unrecognized `subagent_*` payloads pass through as ordinary visible output; only a confirmed subagent context is ever dropped.
- Tests pin the captured wire shapes and the suppression/merge rules; one pipeline test proves the started row survives history filtering.

## Visual explanation

```mermaid
sequenceDiagram
    participant D as devin acp
    participant C as AgentClient sessionUpdate
    participant H as history pipeline
    participant A as history applier

    C->>D: initialize with subagentSupport capability
    D->>C: tool_call_update with subagent_started meta
    C->>C: record taskId
    C->>H: lifecycle row survives filter exemption
    H->>A: materialize subagent_task in progress
    D->>C: chunks/usage tagged subagent_context
    C->>C: drop non-tool internals at ingress
    D->>C: tagged tool_call / tool_call_update
    C->>H: flow through for edit evidence
    H->>A: merge into existing permission row, else drop
    D->>C: tool_call_update with subagent_completed meta
    H->>A: merge task completed with summary
```

## Before / after

| Before | After |
| ------ | ----- |
| Subagent appears as a `run_subagent` tool call that completes instantly with "Background subagent started." | Subagent appears as a task row: in_progress on start (profile, model, description), completed/failed with the subagent's result summary |
| Subagent's internal chunks/tool calls never sent (degraded mode) | Sent with `subagent_context` tags; non-tool internals dropped at ingress, tool updates suppressed in the applier unless they resolve a permission-written row |
| Started row would be lost to snapshot compaction | Lifecycle rows exempted like `_meta.lody.task` snapshots |

## Test plan

- `pnpm vitest run tests/devin-subagent-task.test.ts tests/acp-history-apply.test.ts` in `packages/shared` — 53 passed.
- `pnpm vitest run tests/acp-devin-subagent.test.ts tests/acp-plan-sync.test.ts tests/acp-history.test.ts tests/acp-apply-notification-on-history.test.ts tests/acp-history-batching-equivalence.test.ts tests/claude-code-raw.test.ts` in `apps/cli` — 80 passed.
- `pnpm typecheck` clean in `packages/shared` and `apps/cli`; `pnpm run docs check` reports no errors.
- Verified the wire shapes and capability negotiation against a live `devin acp` 3000.10.21 session during development; not re-run end-to-end in the Lody UI.

- **Review focus:** the ingress drop in `sessionUpdate` (before `handleUsageUpdate`, after `lastSessionUpdateAtMs`), the applier suppression rule ordering (before history mutations, after `primeToolCallIndexForBatch`), and the `filterNotificationsForHistory` exemption.
- **Decisions to challenge:** dropping non-tool internals at ingress rather than marking and gating each consumer (chosen because the consumer list is not enumerable); letting tagged tool updates flow so permission rows and edit evidence resolve (accepted cost: a permission-requested subagent tool shows as a flat `tool_call` row with no task parentage); `"root"` treated as a sentinel, never a task id.
- **Plausible failures / evidence gaps:** the `cognition.ai/*` protocol is private and undocumented; fail-open covers unknown owner ids and unrecognized markers, but semantic drift under an already-known parent id could still classify a new-shape nested lifecycle row as internal. Only `subagentSupport` is advertised (no `subagentControl`, so cancel/foreground remain unnegotiated — the task panel cancel affordance is a pre-existing gap shared with Claude/Codex, not introduced here).

### Authoring context

- **User goal / directives:** Devin CLI (registry agent) subagents should render as task lifecycle entries with their result, matching the Codex/Claude task panel, via the smallest provider-scoped change; nested transcript rendering and subagent control methods are explicitly deferred.
- **Constraints / non-goals:** no changes to the provider-neutral `SubagentTaskPayload` schema, no transcript nesting UI, no `subagentControl` methods; other providers' paths untouched.
- **Risk-bearing decisions:** capability advertisement is gated on `agentType === 'devin'`; parsers fail closed to null so private-protocol drift degrades to ordinary tool calls rather than breaking history application; suppression is doubly gated (known owner id at ingress, materialized task row in the applier) so neither layer alone can silently lose output.
- **Destructive or irreversible behavior:** none; subagent-internal updates are excluded from persisted history by design (they were never persisted before this change either).
- **Deliberately not done or tested:** no live Lody-UI end-to-end run against `devin acp`; `subagents/list`/cancel/foreground left unnegotiated; the task panel cancel affordance mismatch (machine-level vs agent-level capability) is a pre-existing gap filed for follow-up, not fixed here.
- **Unknowns / confidence:** moderate — field names/values were verified against Devin CLI 3000.10.21, but Cognition may change the private protocol; the failure mode in that case is graceful degradation to visible output rather than data loss.

### Original user prompt

<details>

````text
啊 ok，这个稳定性如何？我发现 subagent 在 Lody 里不会正确显示，是正常的吗？是 Lody 的 bug 还是 Devin 的问题？比如 use lody mcp to query session[id: SESSION_ID_REDACTED] history 这个里就比较明显

你抓一下呢，我们本来就有这个 Devin CLI 来接。走的不是 Devin Desktop

可以啊，那我们开始写分支，准备这个 PR。你看一下仓库的 contributing MD 对于贡献还有什么别的要求吗？
````

</details>


Astro-Han:
Ready for review 🙂 @zxch3n — this builds on the Devin CLI integration you landed in #763 and touches the shared ACP history pipeline; the wire shapes were verified against a real `devin acp` capture. @Leeeon233 — heads up on the small shared change in `claude-subagent-task.ts`: `mergeSubagentTaskPayload` now keeps terminal status monotone so a replayed `task_started` can't resurrect a finished task (affects all providers; reasoning in the Agent Note). Happy to adjust anything.

zxch3n:
Awesome! Thanks @Astro-Han

PR #784, opened 2026-09-17
fix: reopen exact child session from creation cards

Leeeon233:
## Problem / pressure

Created-session cards could not reliably reopen a child conversation after its tab was closed. Routing through the child root lost the reopen intent, and selecting immediately after the reopen write could race the local metadata projection, causing the close reconciler to show an open sibling instead.

## Summary

- Resolve bare created-session targets against the mounted Session workspace.
- Reopen closed or historically archived child tabs through the existing lifecycle action.
- Wait for the exact target open state to project locally before selecting it.
- Cancel a pending selection when the user navigates elsewhere.
- Document the closure guarantee and cover the A/B/C sibling race with regression tests.

## Visual explanation

```mermaid
flowchart LR
  Card[View session: child C] --> Known{Known local tab?}
  Known -- No --> Route[Normal Session route]
  Known -- Yes, open --> Select[Select exact child C]
  Known -- Yes, closed --> Reopen[Write reopen for child C]
  Reopen --> Project{C open state projected?}
  Project -- No --> Wait[Keep exact C request pending]
  Wait --> Project
  Project -- Yes --> Select
  Select --> URL[URL becomes session:child-C]
```

## Before / after

| Before | After |
| ------ | ----- |
| A closed child card routed through the child root and could remain closed. | The card restores the known child tab inside the mounted workspace. |
| A stale close projection could redirect explicit child C selection to open sibling A or B. | Selection waits for C to be locally open, then targets C exactly. |
| A delayed restore could override newer navigation. | A newer user navigation cancels the pending selection. |

## Test plan

- `pnpm check`
- `pnpm run docs check`
- `pnpm format`
- `pnpm --filter @lody/components exec vitest run tests/session-navigation.test.ts tests/use-session-actions.test.ts tests/session-tab-url.test.ts tests/session-tab-bar.test.tsx` (101 tests)


Leeeon233:
P1: pending tab restores can cross Session boundaries

`pendingTabRestoreNavigation` survives the render-phase `sessionId` reset, while cancellation compares only `urlTab`.

Reproduction:
1. Open Session A externally on its parent so `?tab` is absent.
2. Click a created-session card for a closed child.
3. Before the reopen projection completes, switch to Session B whose `?tab` is also absent.

Both source and current tab values remain `undefined`. Once A's restore finishes, B's closed-id set does not contain A's child, so the effect calls B's `navigateToSessionTab` with the stale child id and replaces the user's navigation with a nonexistent pending tab.

Please key the pending request by its source `sessionId` and cancel on mismatch, or invalidate it during the Session reset, with a regression test for the cross-Session/tabless case.

Location: `packages/components/src/components/sessions/session-detail.tsx:2434`

---

Resolved in `28d8e589af2666a31baa2b35d3e11d5727f737f2`. The pending request now carries its source Session id, cancellation checks the current Session, and the tabless cross-Session regression is covered. Independently re-verified: 102 focused tests, component typecheck, docs check, and diff check pass.

PR #786, opened 2026-09-17
fix: use Repo snapshots for session lifecycle targets

### LINKED ISSUE #574: [Bug] Archive can miss child Tabs before metadata cache hydration
### Affected area

Session / workspace / local sync

### Installation method

Built from source

### Lody version or commit

b4d544394d30

### Operating system

macOS 26 arm64; the affected Desktop state flow is platform-independent

### Agent or runtime

Desktop local runtime

### What happened?

On a cold start, Session Detail becomes interactive as soon as the requested root Session is present in the bootstrap metadata cache. The full workspace metadata scan can still be incomplete at that point. If a user immediately archives the root Session, archive target selection reads the partial `sessionMetaCacheAtom` and can miss a child Tab whose `parentSessionId` points to that root.

The result is inconsistent archive state: the root is archived while its lifecycle-owned child Tab remains active.

### What did you expect?

Archiving a root Session should not start until the metadata needed to identify all direct child Tabs is hydrated, or archive target selection should read from a complete source. The root and every direct `parentSessionId` child must transition together.

### How can we reproduce it?

1. Create a root Session A and a child Tab T with `T.parentSessionId = A`.
2. Cold-start Desktop and arrange for bootstrap metadata to expose A before the full workspace metadata scan exposes T.
3. Open Session Detail for A; it resolves because A is already in the active Session cache even though `docMetaCacheReadyAtom` is false.
4. Immediately choose Archive.
5. Observe that archive target selection sees only A, so A becomes archived while T remains active.

### How often does it happen?

Sometimes; it depends on user timing during cold-start metadata hydration.

### Relevant log output

No log output is required. The race follows from Session Detail allowing actions before `docMetaCacheReadyAtom` and archive target selection reading `sessionMetaCacheAtom`.

### Additional context

Discovered while narrowing #569 to the archive-only acceptance criteria from #531. This is pre-existing rather than introduced by #569, but it needs an explicit readiness or complete-query contract and regression coverage.

### Before submitting

wibus-wee:
## Problem / pressure

Session lifecycle operations can read a lagging UI metadata projection even after the underlying Repo has the complete session graph. Archive, restore, and archived-root deletion can therefore miss related Sessions or mutate from an incomplete view.

Fixes #574 

## Summary

- Add a shared Repo snapshot reader that derives canonical Session ids from room ids, excludes deletion markers, validates the root, and applies the existing archive/restore/delete relation rules.
- Bind renderer operations to the selected metadata transport’s initial synchronization and fence writes against runtime disposal or workspace switches.
- Make CLI archive, restore, and delete synchronize before discovery and retain ordered, non-transactional cleanup semantics.
- Update the draft relation Spec, implementation notes, and focused behavioral coverage.

## Visual explanation

```text
Session operation
      |
      v
selected metadata source ready
      |
      v
one Repo.listDoc snapshot
      |
      +--> validate live root
      +--> derive archive descendants or direct children
      |
      v
runtime/workspace still current?
      |
      v
ordered writes + best-effort cleanup + sync confirmation
```

## Before / after

| Before | After |
| --- | --- |
| Renderer actions used a UI cache or mixed Repo/cache reads. | All relation-discovering actions use one ready Repo snapshot. |
| Restore could run against incomplete projection state. | Restore and deletion validate and discover from the same snapshot. |
| CLI synchronization and discovery were split across operation helpers. | CLI synchronizes before discovery, applies the snapshot target set, then confirms synchronization. |

## Test plan

- `git diff --check`: passed.
- `pnpm run docs status`: run at start; the changed Spec and bilingual Agent Note are current.
- `pnpm run docs check`: blocked by the repository’s existing missing ACP-submodule links (28 baseline broken links).
- `pnpm check` and `pnpm format`: blocked in this nested worktree because dependencies are absent and Node 26 has no `corepack` shim.
- The implementation Note records focused component/CLI/runtime suites and repository validation completed in a dependency-initialized clone at the same revision.


wibus-wee:
@copilot+gpt-5.6-luna fix static checks eror

